PR #1, opened 2021-04-05
Restore static and media ROOT and URL.

DinoBektesevic:
This is required for file hosting with Django. Everything else just accidentally works in debug mode.

I'm not sure we need `ADMIN_MEDIA_PREFIX` but I guess if we ever build tools for admin we might. In any case it should probably point away from `/media/` but isn't a problem now.


mrawls (approved):
You found and fixed my bug - thank you!!

PR #2, opened 2021-04-05
Add a more flexible and secure authentication and security critical configuration handling.

DinoBektesevic:
Added a `Config` class that reads things from a YAML config file. This is more flexible and more practical going forward. 

Should be improved perhaps to better mask security critical information. The problem is that it's long lived and globally accessible. Will do for now.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,6 @@
 asgiref==3.3.1
 astropy==4.2
+boto3==1.17.59
 certifi==2020.12.5
 click==7.1.2
 cycler==0.10.0
@@ -11,6 +12,7 @@ Jinja2==2.11.3
 kiwisolver==1.3.1
 MarkupSafe==1.1.1
 matplotlib==3.3.4
+moto==2.0.5
 numpy==1.20.1
 Pillow==8.1.2
 pyerfa==1.7.2
--- a/trail/trail/config.py
+++ b/trail/trail/config.py
@@ -12,7 +12,7 @@ CONF_FILE_ENVVAR = "TARILBLAZER_CONFIG"
 configuration file. When the env var does not exist, configuration is assumed
 to exist at its default location (see ``CONF_FILE_PATH``)."""
 
-CONF_FILE_PATH = "~/.trail/config.yaml"
+CONF_FILE_PATH = "~/.trail/secrets.yaml"
 """Default path at which it is expected the config file can be found. Will be
 ignored if ``CONF_FILE_ENVVAR`` env var exists."""
 
@@ -26,6 +26,8 @@ class Config():
     confDict : `dict`
         Dictionary whose keys will be mapped to attributes of the class.
     """
+    configKey = "*"
+
     def __init__(self, confDict):
         self._keys = []
         self._subConfs = []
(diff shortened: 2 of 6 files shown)


spenczar:
Just a drive-by comment: idiomatic Django would be to put all configuration in `trail/settings.py`. You _can_ do your own thing, but you may have trouble using off-the-shelf Django extensions, which will certainly all expect you to put things in settings.py.


What is the advantage of putting config into a `yaml` file?

DinoBektesevic:
I can't exactly put PostgreSQL connection details into `settings.py` without manually sanitizing the file before every commit. Same goes for Django's `SECRET_KEY` This invites problems, no?

The config class is there to punt out the security critical information out of the regularly committed code and into an external, not committed, file. 

As for why YAML: I could continue with having my postgresql data in one config (I don't know, maybe even the `.pgpass` files would work, haven't tried)l, then the Django secret key in another, then probably we will want to record some name of a bucket or a tidbit or two of S3 data somewhere and YAML makes that quick and easy to do from a single file while, down the line, leaving the option of splitting this up into multiple files if I want.

spenczar:
Okay, got it - that makes sense. A few suggestions.

I think it makes sense to rename this to something like `secrets.yaml`, and only put secret values into it. That will help clarify intent, and help ensure that settings appear in fewer places. The current code includes setting stuff like "static_root: ~/trail/static" which seem superfluous; you may as well put that in `settings.py`.

A separate question is whether `yaml` files are the right way to store secrets. I think this is way looser, but I can just post some options for you to consider.

1. Environment variables - this is probably the most common way people do this in Django. It's how a system like Heroku would expect to work, it's how Github Actions CI would like to pass secrets in, it's easy to do in containers, etc. Your server's run script needs to set the environment variables, then.
2. Secrets managers - this is a little fancier but can be nice: you can store secrets in something like AWS secrets manager, and then retrieve them at runtime. For example, [`scimma-admin` defines a `get_secret` function](https://github.com/scimma/scimma-admin/blob/9892fbc1eeb3f8c7c0e769574f52c2cfbf40601a/scimma_admin/scimma_admin/settings.py#L19-L21) which is used to [get the DB password](https://github.com/scimma/scimma-admin/blob/9892fbc1eeb3f8c7c0e769574f52c2cfbf40601a/scimma_admin/scimma_admin/settings.py#L142)
3. Some people make yet another `.py` file with production-specific settings, and then do `from prod_settings import *`; the import statement overrides names with production versions. Then, `prod_settings` is a file that you don't commit - it only exists in your deployed server. I kinda hate this, I think that `import` has too much confusing behavior. But it's common.

Reading secrets from a file, though, is just fine too if that's what you prefer.

DinoBektesevic:
Hello @spenczar - sorry for the long delay, between various obligations I quite **literally** could not make any significant progress in last two weeks before this weekend. 

I took to heart your recommendations and renamed the class, I added, and tested in deployment, functionality for secrets.       
I have not added env var overrides today because I wanted to focus more on the processing part, but it should be trivial to add specific class overrides now too. 
Hope this ticks off at least some your checkboxes.

mrawls:
Let's close this out so we're not blocking #3 anymore

mrawls (approved):

Issue #9, opened 2021-04-05
Build a CI.


DinoBektesevic:
Currently of little priority as at best we have 1 test in the package but a good reminder to implement some and get GA to run them.

mrawls:
Adding a note so we remember - we need to get some kind of auto-linting set up soon.

PR #10, opened 2021-04-07
Standardize header metadata.

### LINKED ISSUE #7: Header metadata and DB schema definitions
The biggest blocker to moving forward in building out the query and image metadata handling is the fact that there are no explicit decisions on what header metadata we want to, or even can, keep. 

I assume it's safe to assume that if we use Astropy WCS class we can reliably define a standard interface to standard WCS information for a wide array of instruments. The remaining keywords are there for some, not there for others. 

Even a provisional list of header keywords we want to track and standardize on is a start, even if those keywords are not always present for all instruments, but without at least that I can't map any header to a table schema and I certainly have a much harder time designing a query interface.

DinoBektesevic:
This is a first attempt at taking the various varied FITS headers and standardizing a select number of header keywords that we can use in our header metadata DB table.

The header keywords we standardize on here will set our table schema and our query tool interface but there are some outstanding issues with standardizing the WCSs. 
If we can not easily separate WCS data into a standardized set of columns querying that table will be much harder but it should still be possible to save a subset of keywords (center pixel values and coordinates) and the whole WCS as a pickled blob. 

This would not be optimal, as optimal as storing all of these values separately,  I suspect. Ergo, this draft PR to give some insight into what the difficulties are.


trail/upload/models.py
```diff
@@ -3,29 +3,29 @@
...
+    # verbose_names should be lowercase, Django will capitalize
+    # https://docs.djangoproject.com/en/3.1/topics/db/models/#verbose-field-names
+    name = models.CharField("name of used translator.", max_length=10, null=False)
+    instrument = models.CharField("instrument name", max_length=20)
+    telescope = models.CharField("telescope", max_length=20, null=False)
+    science_program = models.CharField("science program image is a part of.", max_length=30)
+
+    obs_lon = models.FloatField("observatory longitude (deg)", null=False)
+    obs_lat = models.FloatField("observatory latitude (deg)", null=False)
+    obs_height = models.FloatField("observatory ASL height (m)", null=False)
```
mrawls: Not sure if this is legitimately "above sea level," let's just call it "observatory height (m)"

trail/upload/models.py
```diff
@@ -3,29 +3,29 @@
...
+
+    datetime_begin = models.DateTimeField("UTC at exposure start.", null=False)
+    datetime_end = models.DateTimeField("UTC at exposure end.", null=False)
+
+    radius = models.FloatField("distance between center and corner pixel", null=False)
```
mrawls: prefix this and all the following things with wcs, e.g., `wcs_radius`

trail/upload/process_uploads/processors.py
```diff
@@ -0,0 +1,286 @@
...
+            location = translated.to_location()
+
+            standardizedKeys["obs_lon"] = location.lon.value
+            standardizedKeys["obs_lat"] = location.lat.value
+            standardizedKeys["obs_height"] = location.height.value
+
+            # translator names are easy everyday names
+            # the rest I don't know how they'll vary over different
+            # instruments but I can assume potentially by a lot?
+            standardizedKeys["name"] = translated.name
```
mrawls: Call this `metadata_translator_name` or something a bit more specific so people don't misconstrue it as, e.g., the name of a satellite or other target in the image

mrawls:
Metadata still missing: band or filter, exposure duration (grab independently instead of doing end - start if feasible), some kind of processed/reduced flag (probably either "yes" or "unknown")

DinoBektesevic:
> Metadata still missing: band or filter, exposure duration (grab independently instead of doing end - start if feasible), some kind of processed/reduced flag (probably either "yes" or "unknown")

Added filter and exposure time for `astro_metadata_translator` recognized instruments. 

Added test dataset to get an overview of where functionality is right now. Coverage right now is about 50/50 of the dataset you compiled. Images in the dataset have been cropped for size but the WCS values have been shifted to maintain positional accuracy (approximately). 

I started gnawing at getting more instruments supported today but nothing that I would like to commit quite yet. I think our discussion was on-point and that whether we want or no we will have to create a `HeaderStandardizer` and a `ImageStandardizer` classes that do the atomic work on a single header and image and then see if `FitsProcessor` can be made to support both single and multi-extension fits files. If not a bit of a redesign of the `FitsProcessor` into a `SingleExtensionFitsProcessor` and a `MultiExtensionFitsProcessor` might be needed so I don't want to promote this into a full PR quite yet.

DinoBektesevic:
Ok, this is now getting much better. I think someone may take a look at what is in this code and not feel offended. 

Added header standardizer classes, essentially maps between what we want in our DB and what we can find in particular instrument's header

Processors are essentially recipes of how particular file should be processed. For example, a compressed archive needs to be uncompressed, "unarchived" then each file needs to be processed; but is this particular file Rubin calexp - then the image is only in 1st HDU - or is it a DECam NOAO image - then all HDUs have headers and data - is it an SDSS.... 



To do:
* rework the DB schema to be able to accept focal planes.        
    * Different pipelines produce different types of FITS files. Some store all CCDs from a focal plane in the same FITS, some split them up. The fake DB model right now accepts everything as a single CCD. We either unravel multiext fits into individual CCDs on a generic basis or we implement a better schema. 
    * DB schema could also use a lot of work in terms of recording uploading times, ips, etc...
*  Fix the `astro_metadata_translator` issues with `Header` https://github.com/lsst/astro_metadata_translator/pull/53
* Fix science validity of the parsed headers.     
    * As-is we are not parsing MOA-II fits file WCS data correctly and it's unclear how to. Perhaps this is a good test case for Astrometry.NET integration. Other headers insert data into the DB, but timezones and other details could be an issue. 
    * Fix timezone-naive `Time.iso` output.     
Format for the time-zone aware `datetime` objects seem to be a bit different and Astropy does not seem to return timezone encoded strings.
    * Standardize database schema values, as is instruments, telescopes, filters etc all kind of return something different. For certain things we can guarantee consistency (f.e. what is in telescope, what in instrument etc. for others, i.e. filter or science-program, we can do our best to be as close to something standard but really can not be consistent). Perhaps make these fields something like "identifier" and then adopt the instruments conventional naming scheme (f.e. `run-calcol-filter-field` for SDSS or `run-field-filter-chip` for MOA-II etc.)
    * This is literally an easy free-for-all problem, it just takes emailing people in cases like MOA-II or, for bigger instruments, reading documentation. Las Cumbres standardizer even has a comment to paper with the lookup table.
* Figure out what I want to do with the dataclasses
    * they are very helpful when debugging etc and it is nice when you don't have to traverse highly nested dicts, but their function would ideally be handled by the model classes. Unfortunately the schema is not very reflective of processing steps and will complain for mandatory values for example.
    * no tests for now, until we decide we like or don't like them
* Add a `register` method to processors and standardizers? 
    * gets rid of the import funkery that has to happen right now.
* Add a (manual, or automatic) priority levels to processors/standardizers
    * or remember procs./stand. that volunteered and try to reprocess on fail....
* better naming than `process_upload.upload_processor.process` etc...
    * help appreciated, I am not imaginative
* Consider getting rid of `from*` and `get*` methods on these classes and make instantiation do that bit of work too (makes it prettier?)
* Perhaps a mixin class to standardize all these similar method names and reduce amount of code duplication.
* flakify
* flesh out the tests
    *  as is we essentially just run the entire pipeline end-to-end so we excercise the code, but we should really make it more atomic
* fix how gallery works before this is merged.

trail/upload/process_uploads/processors/decam_processor.py
```diff
@@ -0,0 +1,360 @@
...
+    9:  {"indices": ( (0, 8),  (1, 8),  (2, 8),  (3, 8), (4, 8), (5, 8)),
+         "names":   ( "N19",   "N18",   "N17",   "N16",  "N15",  "N14"),"rtype": "odd"},
+    10: {"indices": ( (1, 9),  (2, 9),  (3, 9),  (4, 9), (5, 9)),
+         "names":   ( "N24",   "N23",   "N22",   "N21",  "N20"), "rtype": "even"},
+    11: {"indices": ((1, 10), (2, 10), (3, 10), (4, 10)),
+         "names":   ( "N28",   "N27",   "N26",   "N25"), "rtype": "odd"},
+    12: {"indices": ((2, 11), (3, 11), (4, 11)),
+         "names":   ( "N31",   "N30",   "N29"), "rtype": "even"},
+}
+"""A row-based layour of the DECam focal plane science detectors."""
```
mrawls, line 51: Double check this against the actual layout at, e.g., http://www.ctio.noao.edu/noao/node/2250. The preview you showed today looked like some things weren't in the right spot, but it's a very nice plot thumbnail for a multi-extension-FITS file.

trail/upload/models.py
```diff
@@ -3,29 +3,31 @@
...
+    science_program = models.CharField("science program image is a part of.", max_length=30)
+
+    obs_lon = models.FloatField("observatory longitude (deg)", null=False)
+    obs_lat = models.FloatField("observatory latitude (deg)", null=False)
+    obs_height = models.FloatField("observatory height (m)", null=False)
+
+    datetime_begin = models.DateTimeField("UTC at exposure start.", null=False)
+    datetime_end = models.DateTimeField("UTC at exposure end.", null=False)
+    exposure_duration = models.FloatField("exposure time (s)")
+    physical_filter = models.CharField("physical filter", max_length=30)
```
mrawls: Consider renaming to `filter_name` since this isn't specifically the physical piece of glass a la Rubin Science Pipelines.

mrawls: Still prefer `filter_name` if you're OK with that

trail/upload/models.py
```diff
@@ -3,29 +3,49 @@
...
+
+
+class Metadata(models.Model):
+
+    # This will need to be fixed, cascading can orphan medata entries
+    upload_info = models.ForeignKey(UploadInfo, on_delete=models.PROTECT, null=False)
+
+    # verbose_names should be lowercase, Django will capitalize
+    # https://docs.djangoproject.com/en/3.1/topics/db/models/#verbose-field-names
+    processor_name = models.CharField("name of used translator.", max_length=20, null=False)
```
mrawls: Please standardize the verbose names without periods, unless you think they enhance clarity. This one in particular could be a bit clearer as, e.g., "name of translator used to process FITS header information"

trail/upload/models.py
```diff
@@ -3,29 +3,49 @@
...
+    standardizer_name = models.CharField("name of used translator.", max_length=20, null=False)
+    instrument = models.CharField("instrument name", max_length=20)
+    telescope = models.CharField("telescope", max_length=20, null=False)
+    science_program = models.CharField("science program image is a part of.", max_length=30)
```
mrawls: Is there any way to indicate a field is required or not? I think "science program" is a nice to have, not an absolutely must have.

trail/upload/models.py
```diff
@@ -3,29 +3,49 @@
...
+class UploadInfo(models.Model):
+    # integer autoincrement primary fields is automatically added by Django
+
+    created = models.DateTimeField("date on which data was uploaded", auto_now_add=True)
+    ip = models.GenericIPAddressField("IP address of the uploader", default="127.0.0.1")
+
+
+class Metadata(models.Model):
+
+    # This will need to be fixed, cascading can orphan medata entries
```
mrawls: Typo in comment (metadata entries, presumably). If this isn't a straightforward fix to do now, please open an issue.

trail/upload/process_uploads/processors/decam_processor.py
mrawls: Did you have a chance to double check this now puts the CCDs in the correct spot?

trail/upload/process_uploads/processors/rubin_processor.py
```diff
@@ -0,0 +1,42 @@
+"""
+Class for processing calibrated FITS files produced by Vera C. Rubin Science
+Pipelines.
+
+Images processed by Rubin do *not* store the entire focal plane worth of data
+in their FITS files, but only a single CCD image, alongside a mask and a
+variance plane data.
+
+We don't want to process variance or mask planes.
```
mrawls, line 9: ... for now 😎

trail/upload/process_uploads/standardized_dataclasses.py
```diff
@@ -0,0 +1,208 @@
...
+__all__ = ["StandardizedWcsKeys",
+           "StandardizedMetadataKeys",
+           "StandardizedHeaderKeys",]
+
+
+dtype = np.dtype([("names", str, 30), ("types", object),  ("defaults", object),
+                  ("required", bool),("closeEquality", bool)])
+
+wcsAttributes = np.array([
+    ("wcs_radius",   float, None, True, True),
```
mrawls: The formatting here doesn't feel sufficiently PEP-8-ey to me, but I could be wrong. I saw one of your commits was "flakeify" so I'm assuming you ran some linter?

trail/upload/process_uploads/standardized_dataclasses.py
```diff
@@ -0,0 +1,208 @@
...
+        return cls(**get_dict_subset(data, cls.keys))
+
+    def values(self):
+        return [getattr(self, key) for key in self.keys]
+
+    def toDict(self):
+        # asdict use here would be nice, but is much slower...
+        return {key: getattr(self, key) for key in self.keys}
+
+    def valid(self):
```
mrawls: Maybe rename this to `isValid` to make it clearer this returns a boolean? You could get *real* bonus points if you used type checking (e.g., `def isValid(self) -> bool:`)

trail/upload/tests/test_processors.py
```diff
@@ -0,0 +1,166 @@
...
+        with open(self.sourceFilePath, "rb") as f:
+            return f.read()
+
+    def temporary_file_path(self):
+        return self.sourceFilePath
+
+
+class TemporaryUploadedFileWrapperTestCase(TestCase):
+    """Tests the TemporaryUploadedFileWrapper functionality. """
+    testDataDir = os.path.join(TESTDIR, "data")
```
mrawls, line 34: Should we consider a separate repository for testdata so the main trailblazer project doesn't get too unwieldy to clone? Or has that ship sailed?

trail/upload/process_uploads/fits_processor.py
```diff
@@ -0,0 +1,213 @@
...
+        # TODO: a note to fix os.path dependency when transitioning to S3
+        # and fix saving method from plt to boto3
+        smallPath = os.path.join(cls.media_root, filename+'_small.jpg')
+        largePath = os.path.join(cls.media_root, filename+'_large.jpg')
+
+        # TODO: consider removing PIL dependency once trail detection is
+        # implemented, if it is implemented via OpenCV
+        normedImage = (normedImage.data*255).astype(np.uint8)
+        # this is grayscale
+        img = Image.fromarray(normedImage, "L")
```
mrawls, line 142: Do we have to use PIL for this? Is it fast or otherwise super nice? If it would be straightforward, I'd love to use the color normalization I mentioned before, which you can find at https://github.com/lsst-dm/ap_pipe-notebooks/blob/master/plotLightcurve.py#L112-L144

trail/upload/process_uploads/fits_processor.py
```diff
@@ -0,0 +1,213 @@
...
+        img = img.resize((basewidth, hsize), Image.ANTIALIAS)
+
+        # img is PIL.Image object - simplify
+        return ({"savepath": largePath, "thumb": normedImage},
+                {"savepath": smallPath, "thumb": np.array(img)})
+
+    @classmethod
+    @abstractmethod
+    def canProcess(cls, uploadedFile, returnHdulist=False):
+        # docstring inherited from baseclass; TODO: check it's True
```
mrawls: I'm a little lost in double-decorator base class land here. What docstring is this supposed to inherit?

trail/upload/process_uploads/fits_processor.py
```diff
@@ -0,0 +1,213 @@
...
+        center and of corner points. The center point is the center of the
+        image as determined by image dimensions, determined directly or
+        via header keywords, and the corner is taken to be the (0,0) pixel.
+        Coordinates are then projected to a unit sphere, and the Cartesian
+        components of the resulting projected points, as well as the distance
+        between the center and corner coordiantes, are calculated.
+        """
+        # All Header operations are done by Standardizers. HeaderStandardizer
+        # handles WCS standardization for all our data, so far.
+        raise NotImplemented()
```
mrawls: I could be wrong, but I _think_ raising `NotImplementedError` is more appropriate than `NotImplemented` here.

trail/upload/process_uploads/standardizers/moa_standardizer.py
```diff
@@ -0,0 +1,85 @@
...
+        standardizedKeys["datetime_end"] = jdend.isoformat()
+
+        standardizedKeys["exposure_duration"] = self.header["EXPTIME"]
+
+        # TODO: filter out what is the filter standardization here
+        standardizedKeys["physical_filter"] = self.header["COLOUR"].strip()
+
+        return standardizedKeys
+
+    def standardizeWcs(self, **kwargs):
```
mrawls, line 71: 😱 do we really need to hard-code this? Two thoughts. Is there some kind of "None" or "nan" for postgres? Or, more at the heart of it, should we just reject the upload outright? Would it help to talk with the folks who sent me this image?

trail/upload/process_uploads/standardizers/moa_standardizer.py
```diff
@@ -0,0 +1,85 @@
...
+        standardizedKeys["telescope"] = self.header["OBSTEL"].strip()
+        standardizedKeys["instrument"] = self.header["CAMERA"].strip()
+
+        run = self.header["RUN"].strip()
+        field = self.header["FIELD"].strip()
+        filter = self.header["COLOUR"].strip()
+        chip = self.header["CHIP"]
+        standardizedKeys["science_program"] = f"{run}-{field}-{filter}-{chip}"
+
+        # TODO: Fix datetimes
```
mrawls, line 38: Oy, time is always hard. Please open an issue for this if you haven't.

trail/upload/process_uploads/header_standardizer.py
```diff
@@ -0,0 +1,329 @@
...
+
+__all__ = ["HeaderStandardizer", ]
+
+
+class HeaderStandardizer(ABC):
+    """Supports standardization of various headers.
+
+    Standardization consists of:
+    * converting WCS information contained in each image-like header such as
+      PrimaryHDU, CompImageHDu etc. (i.e. each header for which it is infered
```
mrawls:
```suggestion
      PrimaryHDU, CompImageHDU etc. (i.e. each header for which it is infered
```

trail/upload/process_uploads/header_standardizer.py
```diff
@@ -0,0 +1,329 @@
...
+    filename : `str`
+        Name of the file from which the HDU was read from, sometimes can encode
+        additional metadata.
+    """
+
+    standardizers = dict()
+    """All registered header standardizers."""
+
+    name = None
+    """Standardizer's name. Only named standardizers will be registered."""
```
mrawls, line 52: Please give me a bit more to go off of in doc-string land. Standardize, translate, and register are all words that mean very specific things around here, but this isn't making it super clear.

trail/upload/process_uploads/header_standardizer.py
```diff
@@ -0,0 +1,329 @@
...
+        The center point is the center of the image as determined its
+        dimensions directly or via header keywords NAXIS1 and NAXIS2, and the
+        corner is taken to be the (0,0)-th pixel.
+        World coords. at the points are to unit sphere, in Cartesian system,
+        and the distance between them is calculated.
+        """
+        standardizedWcs = {}
+        centerX, centerY = int(dimX/2), int(dimY/2)
+
+        # TODO: When eventually logging is added to processing, redirect these
```
mrawls, line 99: This one *definitely* needs an issue opened if we don't have one... I would argue we might initially want to give up if we can't make sense of the header WCS. Maybe later we can get smarter. Won't silently ignoring this lead to a metadata table full of useless WCS-less things?

trail/upload/process_uploads/header_standardizer.py
```diff
@@ -0,0 +1,329 @@
...
+        standardizedWcs["wcs_center_y"] = unitSphereCenter[1]
+        standardizedWcs["wcs_center_z"] = unitSphereCenter[2]
+
+        standardizedWcs["wcs_corner_x"] = unitSphereCorner[0]
+        standardizedWcs["wcs_corner_y"] = unitSphereCorner[1]
+        standardizedWcs["wcs_corner_z"] = unitSphereCorner[2]
+
+        return standardizedWcs
+
+    # wow, do not flip these two decorators around...
```
mrawls, line 140: 👀

trail/upload/process_uploads/header_standardizer.py
```diff
@@ -0,0 +1,329 @@
...
+        header : `object`
+             The header, Astropy HDU and its derivatives.
+        \**kwargs : `dict`
+            Additional keyword arguments
+
+        Keyword arguments
+        -----------------
+        filename : `str`
+            Name of the file from which the HDU was read from, sometimes can encode
+            additional metadata.
```
mrawls, line 187: Truer words were never spoken

trail/upload/process_uploads/header_standardizer.py
```diff
@@ -0,0 +1,329 @@
...
+
+        Returns
+        -------
+        standardizerCls : `cls`
+            Standardizer class that can process the given upload.`
+
+        Raises
+        ------
+        ValueError
+            None of the registered processors can process the  upload.
```
mrawls:
```suggestion
            None of the registered processors can process the upload.
```

trail/upload/process_uploads/header_standardizer.py
```diff
@@ -0,0 +1,329 @@
...
+        """
+        standardizers = []
+        for standardizer in cls.standardizers.values():
+            if standardizer.canStandardize(header):
+                standardizers.append(standardizer)
+        standardizers.sort(key=lambda standardizer: standardizer.priority, reverse=True)
```
mrawls: Any way to rewrite this without a lambda? I just despise them is all.

trail/upload/process_uploads/header_standardizer.py
```diff
@@ -0,0 +1,329 @@
...
+        """Normalizes FITS header information of the primary header unit and
+        returns a dictionary with standardized, as understood by trailblazer,
+        keys.
+
+        Returns
+        -------
+        standardizedKeys : `dict`
+          A dictionary with standardized header keys and values.
+        """
+        raise NotImplemented()
```
mrawls: Clarify somewhere that it's not implemented because it's instrument-specific, please

trail/upload/process_uploads/header_standardizer.py
```diff
@@ -0,0 +1,329 @@
...
+
+    def standardizeWcs(self, hdu=None):
+        """Standardize WCS data a given header. 
+        Standardized keys are the Cartesian components of world coordinates of
+        central and corner points on the image as projected onto a unit sphere
+        and the distance between them.
+
+        Parameters
+        ----------
+        hdu : `obhect` or `None`, optional
```
mrawls:
```suggestion
        hdu : `object` or `None`, optional
```

trail/upload/process_uploads/header_standardizer.py
```diff
@@ -0,0 +1,329 @@
...
+
+        return self._computeStandardizedWcs(header, dimX, dimY)
+
+    def standardize(self, hdu=None):
+        """Convenience function that standardizes the WCS and metadata and
+        returns a dictionary with standardized keys.
+
+        Parameters
+        ----------
+        hdu : `obhect` or `None`, optional
```
mrawls:
```suggestion
        hdu : `object` or `None`, optional
```

trail/upload/process_uploads/header_standardizer.py
```diff
@@ -0,0 +1,329 @@
...
+        centerSkyCoord = wcs.pixel_to_world(centerX, centerY)
+        centerRa = centerSkyCoord.ra.to(u.deg)
+        centerDec = centerSkyCoord.dec.to(u.deg)
+
+        cornerSkyCoord = wcs.pixel_to_world(0, 0)
+        cornerRa = cornerSkyCoord.ra.to(u.deg)
+        cornerDec = cornerSkyCoord.dec.to(u.deg)
+
+        unitSphereCenter = np.array([
+            np.cos(centerDec) * np.cos(centerRa),
```
mrawls, line 116: You've forced all the Ras and Decs to be in degrees, and then done trigonometry. Are you 100% sure that's right?

trail/upload/process_uploads/processors/decam_processor.py
```diff
@@ -0,0 +1,360 @@
...
+        smallPath = os.path.join(self.media_root, self.uploadedFile.basename+'_plane_small.jpg')
+        largePath = os.path.join(self.media_root, self.uploadedFile.basename+'_plane_large.jpg')
+
+        smallThumb = self._createFocalPlaneImage(smallPlane)
+        plt.imsave(smallPath, smallThumb.planeImage.T, pil_kwargs={"quality":30})
+        # due to the size of these images immediately release memory
+        del smallThumb
+
+        largeThumb = self._createFocalPlaneImage(largePlane)
+        plt.imsave(largePath, largeThumb.planeImage.T, pil_kwargs={"quality":30})
```
mrawls: Policy for file/image overwriting error-handling situations?

trail/upload/process_uploads/processors/multi_extension_fits.py
```diff
@@ -0,0 +1,86 @@
...
+
+    @staticmethod
+    def _isImageLikeHDU(hdu):
+        if not any((isinstance(hdu, CompImageHDU), isinstance(hdu, PrimaryHDU),
+                    isinstance(hdu, ImageHDU))):
+            return False
+
+        # People store all kind of stuff even in ImageHDUs, let's make sure we
+        # don't crash the server by saving 120k x 8000k table disguised as an
+        # image (I'm looking at you SDSS!)
```
mrawls, line 40: Good thinking... yikes

trail/upload/process_uploads/processors/single_extension_fits.py
```diff
@@ -0,0 +1,51 @@
...
+        meta = Metadata(upload_info=self.uploadInfo, **header["metadata"])
+        meta.save()
+
+        wcs = Wcs(metadata=meta, **header["wcs"])
+        wcs.save()
+
+    def storeThumbnails(self):
+        large, small = self._createThumbnails(self.uploadedFile.basename, self.imageData)
+        plt.imsave(small["savepath"], small["thumb"], pil_kwargs={"quality": 10})
+        plt.imsave(large["savepath"], large["thumb"], pil_kwargs={"quality": 30})
```
mrawls: Again, question about what to do if this would be overwriting an existing image/file

trail/upload/process_uploads/standardized_dataclasses.py
```diff
@@ -0,0 +1,208 @@
...
+@dataclass(order=True)
+class StandardizedMetadataBase(ABC, StandardizedKeysMixin):
+    keys: ClassVar[list] = metaAttrs["names"]
+    keyTypes: ClassVar[list] = metaAttrs["types"]
+    _closeEqualityKeys = metaAttrs[metaAttrs["closeEquality"] == True]["names"]
+
+    _requiredKeys: ClassVar[list] = metaAttrs[metaAttrs["required"] == True]["names"]
+    _optionalKeys: ClassVar[list] = metaAttrs[metaAttrs["required"] == False]["names"]
+
+    def isCloseTo(self, other, **kwargs):
```
mrawls: Docstring plz, in fact the whole Mixin class could use some more docstrings

mrawls (commented):
That was a marathon of a review! Impressive work, let me know if any of my comments are unclear. I trust you to merge when you think you've addressed them.

trail/upload/models.py
DinoBektesevic: The `null` argument sets what is allowed to be null in the database. There is a more stringent requirement that sets what is required to be present at instantiation but I thought this was sufficient. 

In any case, the standardized dataclasses functionality has now been merged into the models so now we have `keys` and `required_keys` properties that allow inspecting what is required rather easily as well.

trail/upload/models.py
DinoBektesevic: This is me struggling against the Django ORM and schema layouts. Technically I think `CASCADING` would be fine here too, because it works the other way around, but I'm not sure how Django deals with orphaning in reverse so I had these protected. 

Of course, with how we are processing (in steps) we will need to address how we rollback changes and deal with cleanup. This however made thinking about the problem, for me, much harder so I operated under the assumption that all data we get will not fail just to get the code going. 

I think handling failures (both the very explicit ones like raised errors but also more subtle processing errors like badly processed data, or correct data but someone is trolling us with not-science images in data arrays) is really important but it's also confusing. This deserves an issue of its own where rollback and validation behavior is precisely defined. 

A part of the reason why this is so hard is that I'm not even sure what are all the failure cases. Now that we have something, this should definitely get tested out in anger and failures mapped out so that we know what our edge cases are. Maybe someone cleverer than me can reason about it in advance - in which case `FitsProcessor.process` method would be the most appropriate place to start implementing rollback behavior.

For most of the code in this PR I have assumed that no uploaded data will break the code in any non-trivial ways.

trail/upload/process_uploads/standardized_dataclasses.py
DinoBektesevic: It's not, but I dare someone to codify and debug a layout in a non-column-aligned format. My preference here is to ignore it in flake. 

Re: general state of PEP 8 compliance - I _ran_ a linter and it printed stuff out about which I didn't do too much. 
Lintering should probably be sorted out as a part of a CI issue, before it gets out of hand.

trail/upload/process_uploads/standardized_dataclasses.py
DinoBektesevic: We can add typechecking everywhere, or nowhere. I'm not really for a mixed bag though. In any case, I dropped this method when I merged it with models in order to implement it at a later date when we sort out failure cases.

trail/upload/tests/test_processors.py
DinoBektesevic: The image data has been cropped, the WCSs have been translated to fit the crop center and header keys adjusted appropriately. The total size of the dataset is 4MB, most of which in the DECam data (2MB), which could be much much smaller if I leave only a couple of cropped headers in it and make the rest a 1x1 array. If 4MB is too much we can move stuff out, otherwise I'm quite fine with it living here in this state. 

Perhaps if we get some kind of validation stuff going we will want actual full datasets at which point additional resources might be required.

trail/upload/process_uploads/fits_processor.py
DinoBektesevic: AFAIK `matplotlib` can't be used to resize an image, so we need something else. Matplotlib will use PIL in the background anyhow, every time we [save an image ](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.imsave.html). 

PIL here is used just for resizing of the image, not its normalization, because it already comes with `matplotlib`. I guess we can transition to something like scikit image module for this but I wasn't sure if we want to install the entire scikit and scikit image extension for a single operation.
Eventually, if we implement some kind of image science data analysis stuff, I assume we will use a more powerful libraries like OpenCV or something. We can change this to cv resizing method then too. 

Resizing is not particularly slow, although there are magic keywords that can make it even faster. 

As far as image normalization goes: For all cases except DECam focal plane images, and even then only in cases when the file was created by the DECam Community Pipelines, the image normalization is performed by `Astropy.ImageNormalization` (see `FitsProcessor.normalizeImage`). 
We can change the normalization to use the asynh stretch, but that will only work well, for a particular choice of normalization constant, if we're processing calibrated exposures and will often not  be generally applicable, with the same normalization constant, for raw, just isr or fully calibrated image data. 

I picked histogram equalization because it's mathematically guaranteed to maximize contrast (which is what I thought I wanted when we are resizing down this much) but I'm not married to the idea. Again, see the comments in `FiitsProcessor.normalizeImage` method as it was left there as a potential `TODO` for anyone wanting to roll up their sleeves. 

As far as the DECam Community Pipeline image products - I'm not sure we can live with astropy's normalization there, see comments in `DecamFits.normalizeImage` processor method:

>      # astropy equalizer averages 4.3 seconds, the PIL approach can bring
>      # that down to cca 0.13. Without normalization the time is 0.04s

That's with the speedier histogram equalization, performance may take a larger hit.

trail/upload/process_uploads/fits_processor.py
DinoBektesevic: Comment clarified, from `UploadProcessor`.

trail/upload/process_uploads/fits_processor.py
DinoBektesevic: Nice catch! Fixed everywhere where I made this blunder.

trail/upload/process_uploads/standardizers/moa_standardizer.py
DinoBektesevic: It would be great help to hear from people that sent the image to see how they kept track of where in the sky they were looking at because it wasn't clear to me at all. 

For the rest: see the comment about error handling [above](https://github.com/dirac-institute/trailblazer/pull/10#discussion_r654729848)

For cases exactly like this one we might even want to have a "commitable" or "processable" or an actual english word that means we accept or deny files like these - and then we won't have to wait to raise an error and skip wasting cycles on processing.

trail/upload/process_uploads/header_standardizer.py
DinoBektesevic: I've added some clarifying notes sections throughout doc strings for standardizers and processors but I'm open to suggestions.

In short every defined standardizer that sets a `name` will register itself with the top-level standardizer class as an standardizer that is available to standardize a header that meets its requirements in `canProcess`. This is what "registering itself" means. Maybe there's a better expression. The toplevel class dispatches the processing to the standardizer with the highest priority down the line. 

Same goes for processors.

trail/upload/process_uploads/header_standardizer.py
DinoBektesevic: See the comment about error handling [above](https://github.com/dirac-institute/trailblazer/pull/10#discussion_r654729848).

In short: absolutely yes, but it was not wholly clear to me what data is "good enough" to commit and what not.

trail/upload/process_uploads/header_standardizer.py
DinoBektesevic: Sure, it would look something like:

```
# ... for loop ...
def sort_by_priority(standardizer):
    return standardizer.priority
standardizers.sort(key=sort_by_priority, reverse=True)
```

if you're fine with that...

trail/upload/process_uploads/header_standardizer.py
DinoBektesevic: I assume you want this for all abstract base classes methods? Because they will complain quite explicitly when someone tries to define a class without defining all the required abstract methods, I think something like: `Can't instantiate abstract class with abstract methods ...` thing.

trail/upload/process_uploads/header_standardizer.py
DinoBektesevic: Nice catch, thank god for Astropy though because fixing this otherwise would require fixing the test data values too which would be a pain.

```
from astropy import units as u
from astropy.coordinates import SkyCoord
c = SkyCoord(ra=10.625*u.degree, dec=41.2*u.degree, frame='icrs')

c.ra
# <Longitude 10.625 deg>
c.ra.to(u.deg)
# <Longitude 10.625 deg>

import numpy as np
np.cos(c.ra)
# <Quantity 0.98285499>
np.cos(c.ra.to(u.deg))
# <Quantity 0.98285499>
np.cos(c.ra.to(u.rad))
# <Quantity 0.98285499>
```

trail/upload/process_uploads/processors/decam_processor.py
DinoBektesevic: See the comment about error handling [above](https://github.com/dirac-institute/trailblazer/pull/10#discussion_r654729848).

Is overwriting an error or not? I'm not sure in what cases it should be allowed, or should updating data be allowed only by manual cranking it. 
Is processing a file that produces the same standardized metadata an error - our standardized metadata is basically the observer location, observation time and observing instrument so this is not really sufficient to tell (f.e. DECam via Rubin will even have the same observing times but we can't tell they're the same pointing, different ccd at the moment) so we need to check if WCS matches I guess. Then I guess if WCS and metadata matches do we want to soft delete, hard delete or substitute in-place? Anyhow... 

The newest code centralizes the place were we save the images, in preparation for the move to S3. I assume we can make a unique string (read as: randomly generated string large enough to give an asymptotic guarantee of never duplicating) and centralizing that functionality for all processors too, as a quick fix, but there's a good chance this behavior will become clear once we answer above questions so I'm not particularly keen on resolving it in this PR. 

For this PR I've assumed no uploaded file ever will break any of the code in it in non-trivial ways.

trail/upload/process_uploads/processors/single_extension_fits.py
DinoBektesevic: See the comment about error handling [above](https://github.com/dirac-institute/trailblazer/pull/10#discussion_r654734302).

trail/upload/process_uploads/standardized_dataclasses.py
DinoBektesevic: This was merged with models and file trashed, models now have docstrings.

trail/upload/process_uploads/header_standardizer.py
mrawls: I can know that's what it's doing if I stare at it for 10 min. If it doesn't sacrifice speed, I prefer the clarity - thanks 😁

trail/upload/process_uploads/header_standardizer.py
mrawls: Oh that's fair. I meant just in the docstring in these "standard" standardizers, saying explicitly they're ABC methods intended to be overridden (inherited? whatever the right word is) by an instrument-specific one instead. Maybe that's implied by how you wrote it and I'm just not as used to the convention as I should be.

trail/upload/models.py
mrawls: Hmmm... let's discuss more later, because I'd like to better understand the subtleties here. I think this is fine for now, just want to make sure we flag it to revisit when friendly data can't be assumed.

trail/upload/process_uploads/standardized_dataclasses.py
mrawls: I added a comment in #9 about this.

trail/upload/process_uploads/processors/decam_processor.py
mrawls: Yeah I'm not sure either 😄 we need a duplicates policy in general. If all the metadata (including WCS) matches, I think we'd want to default to not accepting the upload so we don't waste our upload/download cloud $ on duplicates. Does that sound reasonable?

trail/upload/models.py
mrawls: I swear I wrote a reply and GitHub ate it. Let's discuss more, because I'd like to understand the subtleties here better. For now, I'm OK with assuming uploaded data is friendly, but we'll need to break that assumption eventually.

trail/upload/process_uploads/fits_processor.py
mrawls: OK great points, PIL it is! I just hadn't seen it before; I guess in Rubin we have all our own image manipulation C++ flavored shenanigans.

mrawls: That's an excellent point about how the normalization I suggested may only work well for halfway calibrated images. We can leave it as histogram equalization to maximize contrast 👍

trail/upload/process_uploads/standardizers/moa_standardizer.py
mrawls: Emailed the source of the MOA image to ask about this just now, we'll see what she says.

trail/upload/process_uploads/standardized_dataclasses.py
mrawls: I'd kind of love to add it everywhere heheheh... it's so nice

trail/upload/models.py
DinoBektesevic: Weird, I see both replies. Ok, issue https://github.com/dirac-institute/trailblazer/issues/15 - this will be added in a separate issue (as much as possible) and we will discuss details at a later date. I will mark discussions like these as resolved here.

trail/upload/process_uploads/fits_processor.py
DinoBektesevic: Yeah, there could be a particular normalization factor for asinh that works well across the board and we could add the normalization factor as an attribute, or a property for those processors that have multiple different appropriate normalization factors, and it would work just fine. It will make the images look nicer. 

Is this resolved, pending an issue creation or should we make this a requirement for this PR?

trail/upload/process_uploads/standardizers/moa_standardizer.py
DinoBektesevic: Issue: https://github.com/dirac-institute/trailblazer/issues/16

trail/upload/process_uploads/header_standardizer.py
DinoBektesevic: No problems with adding the comments, I just was not clear if you meant this one specific method or all abstract methods in the code?

trail/upload/process_uploads/processors/decam_processor.py
DinoBektesevic: This sounds perfectly reasonable to me in this case. I just didn't have the capacity to think about all of the details here. To boot it felt like too much work that will leave more people pending on the merge of this PR.
Punted to issue https://github.com/dirac-institute/trailblazer/issues/15

trail/upload/process_uploads/standardized_dataclasses.py
DinoBektesevic: Issue https://github.com/dirac-institute/trailblazer/issues/17

trail/upload/process_uploads/header_standardizer.py
mrawls: Ideally all of them. Because I'm bad at telling which ones are this flavor vs an actual functionality we haven't implemented yet 🙂 It can be the same generic one line comment.

DinoBektesevic: Ok, will add tonight. None of the methods that currently raise `NotImplementedError` are raising them because it's a thing that isn't actually implemented yet. For some, like MOA, they are fake and should be fixed, for others they actually work. Mostly they raise them because they are meant to be overridden once inherited from the class, at which point they inherit the docstring of the method too. Or so I hope (the docstring bit).

DinoBektesevic:
Ok, fixed the last comments and added tests for Models. 

The remainder of issues here were punted other issues in which we will tackle them. Mainly all is well except: we don't have good error handling, data update/replace/remove logic and MOA-II WCS is a placeholder pending error handling, or getting Astrometry.net solver working.

Looking for any last comments?

mrawls:
Understood. Let's get it merged so we can find fun new ways to break things 😄

PR #11, opened 2021-04-28
html added.

### LINKED ISSUE #5: Improve site design.
The good the bad and the ugly. So far we're missing 2/3rds but I can't do good HTML, CSS nor JS.

eggls6:
Added a little html5 front. Did not use and templates/actual images yet. Open index.html to test.


mrawls:
This is amazing, I love it! I have only very minor requests!

* I don't think the landing page should require scrolling to get to interesting/relevant content (i.e., some "latest" subset that shows a preview of the gallery preview). Once the rest of the apps/pages take shape, I think exactly what belongs on the landing page will become more apparent.
* I was thinking of much smaller gallery thumbnails (perhaps the gallery app will zoom on a small cutout of pixels near the center of the image, or in the future, near the center of a detected streak?) and a whole bunch of them on the full gallery page. More like these: https://sciencephotogallery.com/collections/space, https://www.sciencephoto.com/set/3472/sputnik-images, https://noirlab.edu/public/images/

trail/html/index.html
```diff
@@ -0,0 +1,227 @@
...
+	</head>
+	<body class="is-preload">
+
+		<!-- Wrapper -->
+			<div id="wrapper" class="fade-in">
+
+				<!-- Intro -->
+					<div id="intro">
+						<h1>trailblazer</h1>
+						<p>Your favorite collection of staggering satellite streaks.</p>
```
mrawls, line 23:
```suggestion
						<p>An open data repository for astronomical data products affected by satellites</p>

```
Or, if you want to be snarky/add hover text/footnote, "satellite streaks as a service" :)

eggls6: @Landing: Sure, no problem. Check out the new version! I've also altered the color scheme to UW. Let me know if that works or if you wanna change it. @Gallery: This is going to look nothing like the page at the moment once Dino has hacked in the template, so no worries about that at the moment. I just wanted to put some place holder there. Also, there is no content. Once you know what should go on the various pages, we can revisit.

mrawls:
Perfection. Of course I found another typo, the landing page says "WELCOME TO THE GALERY" with one L (should be GALLERY) 😄 @DinoBektesevic, what do we need to do to get this wonderful html-fest django-ified?

DinoBektesevic:
Thanks @eggls6 this is great! I'll look into translating this to jinja templates as soon as I finish with getting all of the `astrometadata` test data to the point where we can ingest them. It's incredible how non-standard the "standards" are. Regardless, this is great effort and will make our lives easier.

Re: lack of content, there are example images that will be produced by the processing in `trailblazer/trail/media`. This is subject to a bit of change in the future where instead of a single image, each frame will have 2 - one small, for displaying many on the same page, that is easy on the download side and one large - that is the same size as the original frame to which we can link the small gallery preview to. Maybe that helps?

DinoBektesevic:
I lifted the HTML pages and moved them to Django's templates. It's going to look a bit messy in the code diff view because I had to add the assets and images to the trail app so now they'll be duplicated. It should be a bit to see the trees from the forest once we delete html dir. But hopefully if you clone it you can see the following files: 
* in`trail/templates` files `base.html` and `about.html`
* `upload/templates/upload.html`
* gallery/templates/gallery.html`
* `query/templates/query.html`

And how to use the Jinja templating engine to create HTML on the fly. Check, specifically, the gallery template.
This can be improved, mainly by not being so ad-hoc about the layout in the `base.html` template and making better block layouts and block naming.

PR #14, opened 2021-06-18
Clean up apps and directories.

DinoBektesevic:
Cleaning out Siegfrieds HTML directory and the unused blazer app directory.


DinoBektesevic:
@mrawls not sure if you have any vested interest in keeping the "blazer" app around or no.

mrawls:
I'd envisioned that being a thing in the future that identifies streaks or does halfway interesting science/statistics on images! It will remain empty for some time, though, so feel free to toss it for simplicity and we can add it when we actually get to it.

PR #30, opened 2021-09-01
Gallery

### LINKED ISSUE #26: Improve gallery appearance and functionality
The gallery shows thumbnail preview of all uploaded images. Improvements could include:

- Most of what appears on the gallery page ought to be images (not negative space)
- Thumbnail previews should ideally feature a portion of the image containing a satellite streak
- Make it possible for a user to hover or otherwise preview metadata for any image thumbnail
- Add ability to download the FITS files behind the thumbnail preview

Jedijoe100:
Things that this pull request contains:

Import images and their metadata to the gallery page to be displayed.
When clicking an image it redirects to a page for that image with the metadata being displayed.
Adjusted table so that there is less white space.
Added an asynchronous request system for getting new images and loading them onto the main webpage.

Things that probably need to be worked on:

Obtaining data from the database seems inefficient, could probably be better.
Downloading images.

### DIFF
--- a/trail/gallery/templates/gallery.html
+++ b/trail/gallery/templates/gallery.html
@@ -10,9 +10,7 @@
         <header class="major">
             <span class="date">April 25, 2021</span>
             <h2><a href="#">WELCOME TO THE GALLERY</a></h2>
-            <p>Aenean ornare velit lacus varius enim ullamcorper proin aliquam<br/>
-                facilisis ante sed etiam magna interdum congue. Lorem ipsum dolor<br/>
-                amet nullam sed etiam veroeros.</p>
+            <p>Have a scroll through the images and see some images that are affected by satellite trails.</p>
         </header>
         <a href="#" class="image main"><img src="{% static 'images/starlink.jpg' %}" alt=""/></a>
         <ul class="actions special">
@@ -22,8 +20,7 @@
 
     <!--
          Posts
-         All these values These should probably be in a database, and there should be a view that retrieves them
-         and creates the tags.
+         Loads them from the database and currently it is hard coded to display 12 images per page.
     -->
     <section id="image_gallery">
         {% for image in data %}
@@ -36,10 +33,6 @@
                         <span>{{ image.caption }}</span>
                     </div>
                 </a>
-
-                <!--<ul class="actions special">
-                    <li><a href="{% get_media_prefix %}{{ image.name }}" class="button">Full Frame</a></li>
-                </ul>-->
             </div>
         {% endfor %}
     </section>
--- a/trail/gallery/templates/gallery.html
+++ b/trail/gallery/templates/gallery.html
@@ -27,7 +27,7 @@
 <section class="posts">
     {% for image in data %}
     <article>
-		    <a href="{% get_media_prefix %}{{ image.name }}">
+		    <a href="/image?{{ image.id }}">
             <img src="{% get_media_prefix %}{{ image.name }}" width="300"  alt=”{{image.name}}” id=”showSimilarInPopup”>
             <div class=”caption”>
                 <span class="date">{{ image.date }}</span>
--- a/trail/gallery/templates/gallery.html
+++ b/trail/gallery/templates/gallery.html
@@ -13,37 +13,11 @@
             <p>Have a scroll through the images and see some images that are affected by satellite trails.</p>
         </header>
     </article>
-
-    <!--
-         Posts
-         Loads them from the database and currently it is hard coded to display 12 images per page.
-    -->
-    <section id="image_gallery">
-        {% for image in data %}
-            <div class="gallery_item">
-                <a href="/gallery/image?{{ image.id }}">
-                    <img src="{% get_media_prefix %}{{ image.name }}" width="300" alt=”{{ image.name }}”
-                         id=”showSimilarInPopup”>
-                    <!--<div class=”caption”>
-                        <span class="date">{{ image.date }}</span>
-                        <span>{{ image.caption }}</span>
-                    </div>-->
-                </a>
-            </div>
-        {% endfor %}
-    </section>
-
-    <!-- Footer -->
-    <footer>
-        <div class="pagination">
-            <!--<a href="#" class="previous">Prev</a>-->
-            {% for i in num_of_page %}
-                <a id="pagenum_{{ i }}" onclick="get_images({{ i }})" class="page pagenum {% if page == i %} active {% endif %}">{{ i }}</a>
-            {% endfor %}
-        </div>
-    </footer>
+    <div id="table_content">
+        {% include "gallery_table.html" %}
+    </div>
 {% endblock %}
 
 {% block content.scripts.extended %}
-<script src="{% static 'assets/js/pagination_gallery.js' %}"></script>
+<script src="{% static '/assets/js/pagination_gallery.js' %}"></script>
 {% endblock %}
--- a/trail/gallery/templates/gallery.html
+++ b/trail/gallery/templates/gallery.html
@@ -27,17 +27,14 @@
 <section class="posts">
     {% for image in data %}
     <article>
-		    <header>
-			      <span class="date">April, 2017</span>
-			      <h2><a href="#">Sed magna<br />
-			          ipsum faucibus</a></h2>
-		    </header>
 		    <a href="{% get_media_prefix %}{{ image.name }}">
             <img src="{% get_media_prefix %}{{ image.name }}" width="300"  alt=”{{image.name}}” id=”showSimilarInPopup”>
             <div class=”caption”>
+                <span class="date">{{ image.date }}</span>
+                <span>{{image.caption}}</span>
             </div>
         </a>
-		    <p>{{image.caption}}</p>
+
 		    <ul class="actions special">
 			      <li><a href="{% get_media_prefix %}{{ image.name }}" class="button">Full Frame</a></li>
 		    </ul>
--- a/trail/gallery/templates/gallery.html
+++ b/trail/gallery/templates/gallery.html
@@ -21,7 +21,7 @@
     <section id="image_gallery">
         {% for image in data %}
             <div class="gallery_item">
-                <a href="/image?{{ image.id }}">
+                <a href="/gallery/image?{{ image.id }}">
                     <img src="{% get_media_prefix %}{{ image.name }}" width="300" alt=”{{ image.name }}”
                          id=”showSimilarInPopup”>
                     <div class=”caption”>
--- a/trail/gallery/templates/gallery.html
+++ b/trail/gallery/templates/gallery.html
@@ -24,9 +24,9 @@
          All these values These should probably be in a database, and there should be a view that retrieves them
          and creates the tags.
     -->
-    <section class="posts">
+    <section id="image_gallery">
         {% for image in data %}
-            <article>
+            <div class="gallery_item">
                 <a href="/image?{{ image.id }}">
                     <img src="{% get_media_prefix %}{{ image.name }}" width="300" alt=”{{ image.name }}”
                          id=”showSimilarInPopup”>
@@ -36,10 +36,10 @@
                     </div>
                 </a>
 
-                <ul class="actions special">
+                <!--<ul class="actions special">
                     <li><a href="{% get_media_prefix %}{{ image.name }}" class="button">Full Frame</a></li>
-                </ul>
-            </article>
+                </ul>-->
+            </div>
         {% endfor %}
     </section>
 
@@ -48,7 +48,7 @@
         <div class="pagination">
             <!--<a href="#" class="previous">Prev</a>-->
             {% for i in num_of_page %}
-                <a href="/?{{ i }}" class="page {% if page == i %} active {% endif %}">{{ i+1 }}</a>
+                <a href="/?{{ i }}" class="page {% if page == i %} active {% endif %}">{{ i }}</a>
             {% endfor %}
         </div>
     </footer>
--- a/trail/gallery/templates/gallery.html
+++ b/trail/gallery/templates/gallery.html
@@ -12,10 +12,6 @@
             <h2><a href="#">WELCOME TO THE GALLERY</a></h2>
             <p>Have a scroll through the images and see some images that are affected by satellite trails.</p>
         </header>
-        <a href="#" class="image main"><img src="{% static 'images/starlink.jpg' %}" alt=""/></a>
-        <ul class="actions special">
-            <li><a href="#" class="button large">Full Frame</a></li>
-        </ul>
     </article>
 
     <!--
--- a/trail/gallery/templates/gallery.html
+++ b/trail/gallery/templates/gallery.html
@@ -1,59 +1,55 @@
 {% extends "base.html" %}
 {% load static %}
 
-{% block gallery-nav-state%}active{% endblock %}
+{% block gallery-nav-state %}active{% endblock %}
 
 {% block content %}
-<!-- Featured Post -->
-<article class="post featured">
-    <header class="major">
-		    <span class="date">April 25, 2021</span>
-		    <h2><a href="#">WELCOME TO THE GALLERY</a></h2>
-		    <p>Aenean ornare velit lacus varius enim ullamcorper proin aliquam<br />
-				    facilisis ante sed etiam magna interdum congue. Lorem ipsum dolor<br />
-				    amet nullam sed etiam veroeros.</p>
-    </header>
-    <a href="#" class="image main"><img src="{% static 'images/starlink.jpg' %}" alt="" /></a>
-    <ul class="actions special">
-        <li><a href="#" class="button large">Full Frame</a></li>
-    </ul>
-</article>
+    <!-- Featured Post -->
+    <article class="post featured">
+        <header class="major">
+            <span class="date">April 25, 2021</span>
+            <h2><a href="#">WELCOME TO THE GALLERY</a></h2>
+            <p>Aenean ornare velit lacus varius enim ullamcorper proin aliquam<br/>
+                facilisis ante sed etiam magna interdum congue. Lorem ipsum dolor<br/>
+                amet nullam sed etiam veroeros.</p>
+        </header>
+        <a href="#" class="image main"><img src="{% static 'images/starlink.jpg' %}" alt=""/></a>
+        <ul class="actions special">
+            <li><a href="#" class="button large">Full Frame</a></li>
+        </ul>
+    </article>
 
-<!--
-     Posts
-     All these values These should probably be in a database, and there should be a view that retrieves them
-     and creates the tags.
--->
-<section class="posts">
-    {% for image in data %}
-    <article>
-		    <a href="/image?{{ image.id }}">
-            <img src="{% get_media_prefix %}{{ image.name }}" width="300"  alt=”{{image.name}}” id=”showSimilarInPopup”>
-            <div class=”caption”>
-                <span class="date">{{ image.date }}</span>
-                <span>{{image.caption}}</span>
-            </div>
-        </a>
+    <!--
+         Posts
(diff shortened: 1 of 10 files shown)


mrawls:
Excellent work, thank you for the PR!! Initial comments from discussion:
* Dino will do a formal review
* Consider using django pagination, or ensure your custom pagination handles edge cases by adding tests
* Add docstrings to all python functions, you can follow [this guide](https://developer.lsst.io/python/numpydoc.html) for best practices
* Will likely want to move the `image` and `get_image` urls into the main `render_gallery` call

trail/gallery/views.py
```diff
@@ -1,25 +1,93 @@
...
-        return images
-
-
-def render_gallery(request, count=20):
-    images = get_images(count)
-    return render(request, "gallery.html", {'data': images})
+    # Would there maybe a way to sort the data without converting to a list?
+    # converting to a list seems computationally heavy
+    # getting the data from database
+    image_data = list(Thumbnails.objects.values())
```
DinoBektesevic: There is an edgecase that isn't handled: What if there are no thumbnails in the DB? We need to render a clean page. 

RE: your code comments, I looked into it, Django's queries are lazy, so DB will only be hit once the queried object is asked for some data. So a sqlite-like query:
```
SELECT * FROM table LIMIT 5;
```
can be written in Django ORM language as:
```
image_data = list(Thumbnails.objects.values()[:12])
```
See [this part of the manual.](https://docs.djangoproject.com/en/dev/topics/db/queries/#limiting-querysets). 

Note that `values` will force the data to load (i.e. this won't be a lazy load) because it will return a `dict` of values. You should use `all` or `filter` or something else if you want to avoid that, also see next comment.

trail/gallery/views.py
```diff
@@ -1,25 +1,93 @@
...
+
+    # sorting data by date
+
+    # processing data for sending to user
+    # this code seems inefficient, there is probably a less computational heavy approach.
+    for image in image_data:
+        wcs_data = Wcs.objects.values()[image["wcs_id"] - 1]
+        metadata = Metadata.objects.values()[wcs_data["metadata_id"] - 1]
```
DinoBektesevic: There are no guarantees that the ordering of values retrieved from the DB will coincide with their in-DB ordering. Things can be deleted, moved or renamed. You want to resolve the foreign keys by making a lookup in the DB. This can either be another query to the DB or you can write a `JOIN` query.

If you *absolutely* want dates and other caption data, one way of selecting the required fields would be:
```
Thumbnails.objects.values("small",  "wcs__metadata__datetime_begin")
```
No indexing required here. I want to note that if you look at what query this emits you'll see something like:
```
SELECT upload_thumbnails.small, upload_metadata.datetime_begin 
FROM upload_thumbnails 
INNER JOIN upload_wcs ON (upload_thumbnails.wcs_id = upload_wcs.id) 
INNER JOIN upload_metadata ON (upload_wcs.metadata_id = upload_metadata.id)'
```
instead of relying on guessing the `id` values will match their index. 

A drawback to using `values` is that now you have to think about how you get to creating full paths to the thumbnails - because you got a dictionary out of it, not objects. It would be best if you looked at https://github.com/dirac-institute/trailblazer/pull/31 after thinking about it, and come up with a reasonable way to achieve that in the `Thumbnails` if this is the way you decide to go forward.

My personal preference is to kick out any additional information from the gallery (small thumbnail) previews and leave any kind of complicated query making for the image views. Then there already exists functionality on that branch that will do your path creation for you. 

If you are dead-set on having dates and captions in gallery preview, I guess I can live with having an `Thumbnails.make_path` as a classmethod or duplicate some simple data into `Thumbnails` model or something else.

For fetching of full data, in the individual image view case, I'd recommend looking into [`prefetch_related`](https://docs.djangoproject.com/en/3.2/ref/models/querysets/#prefetch-related) that will immediately get all of the related data, and avoid emitting a lot of small individual queries.

trail/gallery/views.py
```diff
@@ -1,25 +1,93 @@
...
+        images.append({"name": image["small"].split("\\media\\")[1],
```
DinoBektesevic: Don't ever split on paths manually, as is - this will only work for POSIX compliant systems. I guess this is a fix around the current path state in the preview DB but it's not advisable to do this in general. 

If you ever find yourself having to do this - this probably means some code somewhere needs to be refactored. In this case this functionality should exist in the place where all things related to thumbnails are - in `Thumbnails`.

trail/gallery/views.py
DinoBektesevic: Second part of the problem here is pagination. You need to work out how to keep prevent always loading the same first 12 images. 

There's two parts to the problem:
First, don't ask for `values` and drop the cast to `list` - this returns a `list(dict)` which is painful to work with. 
Using `all` instead gets you a  `QuerySet` of `Thumbnail` objects which are easier to work with. For example:

```
image_data = Thumbnails.objects.all()[:12]
```
Note this will both be lazy-loading the attributes as well as only selecting the first 12 results. 

Another bonus here is that now you can access the path as `image_data[0].large`. This is where you had problems because paths were absolute in the DB before so you needed to replace them. This should be fixed in https://github.com/dirac-institute/trailblazer/pull/31 .

Do note that if there is some functionality, regarding thumbnails, that is missing here, such as creating absolute paths, or URIs to images or whatever, that functionality should be the prerogative of `Thumbnails` class, not string manipulation.

Secondly, we obviously can't always select only the first 12 objects - think about what kind of arithmetic can you do with `count` and `page` to get to the `start` and `end` indices of your slice and thus avoid having to query for all results. HINT: you already did it couple of lines lower.

trail/gallery/views.py
```diff
@@ -1,25 +1,93 @@
...
+                       "id": image["wcs_id"],
+                       "caption": metadata["telescope"],
+                       "date": metadata["datetime_begin"]})
+    # sorting the images by date with newest images first.
+    images.sort(reverse=True, key=date_sort)
```
DinoBektesevic: You would not need these lines if you don't want dates in the small preview.

trail/gallery/views.py
```diff
@@ -1,25 +1,93 @@
...
+    images = images[page * count:(page + 1) * count]
```
DinoBektesevic: Implement the fix from above (kick out "values" and use `all` and slicing) and you don't need this line at all.

trail/gallery/views.py
```diff
@@ -1,25 +1,93 @@
...
+    return images
+
+
+def render_gallery(request, count=12):
+    """Processes the user request and renders the gallery page
+    or if it is a post request returns information on the next set of images.
+    The input value count is for how many images per request.
+    """
+    number_of_pages = int(np.ceil(len(Thumbnails.objects.values()) / count))
```
DinoBektesevic: This is just a really scary query to execute everytime. Especially because the use of `len` makes the actual query evaluate every time and furthermore the use of `value` makes it non-lazy too.

I have not looked deep enough into the Paginator at the moment to give a clear answer what we should do but at least replace the `len(...` with `Thumbnails.objects.count()` for now. I'm bookmarking this for later, [online blog on thetopic](https://pganalyze.com/blog/pagination-django-postgres), [Django Paginator](https://docs.djangoproject.com/en/3.2/topics/pagination/).

trail/trail/static/assets/js/pagination_gallery.js
```diff
@@ -0,0 +1,49 @@
...
+        },
+        body: page.toString()
+    });//this is the request to the server
+    let data = await response.json()//get the json
+    //Now the code bellow edits the main website changing the images
+    document.getElementById("image_gallery").innerHTML = "";
+    await data.data.forEach(value => {
+        document.getElementById("image_gallery").innerHTML += "<div class=\"gallery_item\">\n" +
+            "                <a href=\"/gallery/image?" + value.id + "\">\n" +
+            "                    <img src=\"/media/" + value.name + "\" width=\"300\" alt=”C:\\Users\\thisi\\PycharmProjects\\trailblazer\\trail\\media\\frame-g-002728-2-0424.png”\n" +
```
DinoBektesevic: We can't be hardcoding paths into our code like this. It would be significantly better if this whole thing was done via the templating engine.

trail/gallery/views.py
```diff
@@ -1,25 +1,93 @@
...
+        elif page < 0:
+            page = 0
+        images = get_images(count, page)
+        return JsonResponse({'data': images})
+
+
+def render_image(request):
+    """Processes the users request and renders the image page
+    """
+    wcs_id = int(request.get_full_path_info().split("?")[1])
```
DinoBektesevic: What happens when there's no question mark? This would be better if it were parsed by `urllib` instead of string manipulation. For example:

```
from urllib.parse import urlparse
parsed = urlparse(request.get_full_path())
wcs_id = parsed.query
```
Is there a specific reason to look at `full_path_info` instead of `full_path`?
I think this raises a good point about uniqueness of the thumbnail naming scheme. I will have to think about that a bit more it seems.

trail/gallery/views.py
```diff
@@ -1,25 +1,93 @@
...
+    image_data = Thumbnails.objects.filter(wcs_id=wcs_id)[0]
+    wcs_data = Wcs.objects.values()[wcs_id - 1]
+    metadata = Metadata.objects.values()[wcs_data["metadata_id"] - 1]
```
DinoBektesevic: All of this should be replaced by a single `value` call using the field traversal scheme like shown earlier, or even, probably, better yet a  `select_related` or `prefetch_related` query to minimize number of times DB is pinged. 

Also, there is no requirement that the ordering of objects retrieved from the DB matches, this is why `JOINS` are used.

trail/trail/static/assets/js/pagination_gallery.js
```diff
@@ -0,0 +1,49 @@
...
+            "                        <span class=\"date\">" + value.date + "</span>\n" +
+            "                        <span>" + value.caption + "</span>\n" +
+            "                    </div>\n" +
+            "                </a>\n" +
+            "            </div>"
+    })
+    for (let i = 0; i < document.getElementsByClassName("pagenum").length; i++) {
+        document.getElementsByClassName("pagenum")[i].classList.remove("active")
+    }
+    document.getElementById("pagenum_" + page).classList.add("active")
```
DinoBektesevic: In fact, all of this is just duplicating code you already have in `views.py`. We should look to remove this because of that. For example, if you use URL parameters, just like you did in the individual image view, you can replace line 41 in `gallery.html` with 

```
<a id="pagenum_{{ i }}" href="/gallery?page={{ i }}" class="page pagenum {% if page == i %} active {% endif %}">{{ i }}</a>
```

add, to the `render_gallery` function in `views.py`, something like 

```
parsed = urlparse(request.get_full_path())
if parsed.query:
    page = int(parsed.query)
else:
    page = 0

[....]
    images = get_images(count, page)
```

and delete all of this plus whatever you have in the `POST` condition in `views.render_gallery` function because it'll never be called.

trail/gallery/urls.py
```diff
@@ -1,11 +1,12 @@
...
 from django.conf.urls.static import static
-from gallery import views
+from gallery.views import render_gallery, render_image
 
 """Route gallery URLs to gallery views here."""
 
 
 urlpatterns = [
-    path('', views.render_gallery, name='gallery'),
+    path('', render_gallery, name='gallery'),
```
DinoBektesevic, line 10: You might want to add

```
path('gallery', render_gallery, name='gallery')
```

in case anyone ever actually clicks on the `gallery` tab and gets sent to `url/gallery`. Not sure what happens now if you do that.

trail/gallery/templates/gallery.html
```diff
@@ -1,62 +1,49 @@
...
-        <a href="#" class="page">3</a>
-        <span class="extra">&hellip;</span>
-        <a href="#" class="page">8</a>
-        <a href="#" class="page">9</a>
-        <a href="#" class="page">10</a>
-        <a href="#" class="next">Next</a>
-    </div>
-</footer>
+{% block content.scripts.extended %}
+<script src="{% static 'assets/js/pagination_gallery.js' %}"></script>
```
DinoBektesevic: See way towards the bottom in the `pagination_gallery.js` and remove this afterwards.

DinoBektesevic (changes requested):
Improvements can be made by removing the JS code. It's completely superfluous and just duplicates what you already have in views. 

I also left some helpful, I hope, tips regarding the code comments left expressing concerns about performance. I think if you check out https://github.com/dirac-institute/trailblazer/pull/31 and think about what kind of functionality you would be missing, or want in addition, to that we can refactor a lot of the path handling out of this code.

Code documentation would be a must and some tests would be great too, but I can add those later if you're not sure how.

Thanks for pointing out the problems with `Thumbnail` model. Should hopefully be a lot better now. I think I should also seriously consider some kind of name fuzzying for saved thumbnails and raw data to avoid clashes in the future.

trail/gallery/urls.py
Jedijoe100: I believe that adding that path is irrelevant because of the code in trail/urls.py which paths this code to both / and /gallery and similarly for /images and/gallery/images.

trail/trail/static/assets/js/pagination_gallery.js
Jedijoe100: The reason this code exists is so that the user can select the next page without having to reload the page every time (that is what the pagination Django code does). I personally find this annoying, however, if the code is too messy or creates more complications than it is worth I can remove it.

DinoBektesevic: Ah, I see... That is a good idea, I should have thought about it more.

I would still like to have a single point where we render the actual layout. It just closest to DRY and makes the easiest version of code to maintain. The fix is pretty simple.

I think the common strategy to solve this is to refactor the template into the gallery and the image-table layout. [Include](https://docs.djangoproject.com/en/3.2/ref/templates/builtins/#include) the image layout template into the gallery one. This mini template now sets out image layout.

Instead of a `JsonResponse` return a rendered HTML snippet (`render(request, 'new_subtemplate' ....)` and then replace the inner HTML of `image_gallery` with whatever text is returned by `render`. Would you be happier with that?

Jedijoe100: Yep, that seems much better than what I am currently doing, will implement it a little bit later.

trail/gallery/templates/gallery_table.html
```diff
@@ -0,0 +1,39 @@
+{% load static %}
+<!--
+         Posts
+         Loads them from the database and currently it is hard coded to display 12 images per page.
+    -->
+<div class="pagination">
+    <!--<a href="#" class="previous">Prev</a>-->
```
DinoBektesevic: If we don't need it best delete it? I'd suggest to remove the comment and resolve this comment.

EDIT: just the `<!--` comment, don't know why it displayed the lines above

trail/gallery/templates/gallery_table.html
```diff
@@ -0,0 +1,39 @@
...
+<section id="image_gallery">
+    {% for image in data %}
+        <div class="gallery_item">
+            <a href="/gallery/image?{{ image.wcs_id }}">
+                <img src="{% get_media_prefix %}{{ image.small }}" width="300" alt=”{{ image.small }}”
+                     id=”showSimilarInPopup”>
+                <!--<div class=”caption”>
+                        <span class="date">{{ image.date }}</span>
+                        <span>{{ image.caption }}</span>
+                    </div>-->
```
DinoBektesevic: Remove.

trail/gallery/templates/gallery_table.html
```diff
@@ -0,0 +1,39 @@
...
+            </a>
+        </div>
+    {% endfor %}
+</section>
+<br>
+<!-- Footer -->
+<footer>
+    <div class="pagination">
+        <!--<a href="#" class="previous">Prev</a>-->
```
DinoBektesevic: Remove.

EDIT: Again, just the `<!-- <a...` comment, don't know why it displayed the lines above

trail/gallery/templates/images.html
```diff
@@ -0,0 +1,26 @@
...
+
+{% block gallery-nav-state%}active{% endblock %}
+
+{% block content %}
+<!-- Featured Post -->
+<article class="post featured">
+    <header class="major">
+		    <span class="date">{{ image_data.datetime_begin}}</span>
+    </header>
+    <!--This will probably need to be changed-->
```
DinoBektesevic: When? Also, if you add `TODO: Change when transition to .... blabla` then one can easily perform a text base search for all `TODO`s, `INFO`, `WARNINGS`, etc...

trail/gallery/tests.py
```diff
@@ -1,3 +1,21 @@
...
+# from gallery.views import get_images
+# from uploads.models import Thumbnails
+#
+#
+# class GalleryImageTestCase(TestCase):
+#     def setUp(self):
+#         objlist = []
+#         for i in range(100):
+#             objlist.append(Thumbnails({"wcs-id": i, "small": "", "large": ""}))
+#         Thumbnails.objects.create(objlist)
```
DinoBektesevic: Try:

```
from django.test import TestCase
from gallery.views import get_images

from django.apps import apps

UploadInfo =  apps.get_model('upload', 'UploadInfo')
Metadata = apps.get_model('upload', 'Metadata')
Thumbnails = apps.get_model('upload', 'Thumbnails')
Wcs = apps.get_model('upload', 'Wcs')

class GalleryImageTestCase(TestCase):
    def setUp(self):
        info = UploadInfo()
        info.save()

        meta = Metadata(
            upload_info=info,
            processor_name="",
            standardizer_name="",
            instrument="",
            telescope="",
            science_program="",
            obs_lon=1.0,
            obs_lat=1.0,
            obs_height=1.0,
            datetime_begin='2009-10-17T01:53:42.1',
            datetime_end='2009-10-18T01:53:42.1',
            exposure_duration=30.0,
            filter_name='i'
        )
        meta.save()

        thumbs = []
        for i in range(100):
            wcs = Wcs(
                metadata = meta,
                wcs_radius=1.0,
                wcs_center_x=1.0,
                wcs_center_y=1.0,
                wcs_center_z=1.0,
                wcs_corner_x=1.0,
                wcs_corner_y=1.0,
                wcs_corner_z=1.0
                )
            wcs.save()
            thumbs.append(Thumbnails(wcs=wcs, small="", large=""))

        Thumbnails.objects.bulk_create(thumbs)

    def testImageListSame(self):
        """test that the pages are working"""
        print(Thumbnails.objects.all().values("wcs"))
```

Unfortunately all relationships on the table have to be respected, even in testing, so without creating the dependencies for the tables to reference you will not be able to insert the `Thumbnails` (depend on `Wcs`, depend on `Metadata`, depends on `UploadInfo`).

Open an issue to remind me to punt this into a function that we can have somewhere in global `test_utils.py`for convenience. Re: the comments in your last commit - you can run tests with `python manage.py test -v 2` for pretty line-by-line output. If you add docstrings to your test class - that's the text that will appear in that output with some dots and an `OK` or a `FAIL` message.

trail/gallery/views.py
```diff
@@ -1,25 +1,71 @@
 from django.shortcuts import render
-from django.conf import settings
-import os
+# from django.conf import settings
+from django.apps import apps
+from urllib.parse import urlparse
+import numpy as np
+
+# import os
```
DinoBektesevic: Remove.

EDIT: again, just the import

trail/gallery/views.py
```diff
@@ -1,25 +1,71 @@
 from django.shortcuts import render
-from django.conf import settings
-import os
+# from django.conf import settings
```
DinoBektesevic: Remove the commented import.

DinoBektesevic (changes requested):

DinoBektesevic (approved):
Thanks a lot for all the work, looks great!
You can go ahead and press the big green button.

PR #31, opened 2021-09-06
Make thumbnail paths in the database relative to some configurable root location.

DinoBektesevic:
Fixes the issue of hardcoding absolute posix paths into the DB. 
Recorded paths are now relative to some root location, this opens the possibility of making a test data repo.


DinoBektesevic:
I think the two "big" problems to solve here, as noticed in , are naming schemas for the thumbnails. Perhaps pre-pending them with the autoincremented wcs-id would work but the problem there is that we might not have resolved the ID at the time we are naming the thumbnails so a random string would be easier and somewhat cleaner (less constraints on when we can commit to DB). Thanks to @Jedijoe100 for pointing this out in https://github.com/dirac-institute/trailblazer/pull/30

I'm happy to hear more opinions. 

Lines in question:

https://github.com/dirac-institute/trailblazer/blob/d9b5175ed968f8b2d48fa3a86564375a8d914007/trail/upload/process_uploads/fits_processor.py#L138 

https://github.com/dirac-institute/trailblazer/blob/d9b5175ed968f8b2d48fa3a86564375a8d914007/trail/upload/process_uploads/processors/decam_processor.py#L344

PR #32, opened 2021-09-06
Unify locations of default configuration.

DinoBektesevic:
Change the settings.py so that path configration occurs on top.
Separate the configuration of locations where we store data from
the static and media location used atm as a crutch.
Add a env var that allows us to point to a different config file than
the default one.
In combination with making all paths in the DB relative ones, this
affords us the opportunity to move our test data to a different repo.

Is technically not blocked by https://github.com/dirac-institute/trailblazer/pull/31 but should be manually tested with https://github.com/dirac-institute/trailblazer/pull/31 and (not yet existent) schema changes to make gallery code smoother before it's made into a full PR.


DinoBektesevic:
This seems to work ok with https://github.com/dirac-institute/trailblazer-testdata repository but should definitely wait for https://github.com/dirac-institute/trailblazer/pull/31 to merge to simplify conflict resolution.

Should add tests for the new methods, but they are very `os` module dependent so it can wait for a more general "URI" class based on `posixpath` and `urllib`.

PR #36, opened 2021-09-17
astronomy.net querying

### LINKED ISSUE #18: Use the Astrometry.net as the WCS solver when WCS information is not present in the header.
Use Astrometry.net as our WCS solver for FITS Headers that contain no WCS information.

Jedijoe100:
Creates a function that sends the fits file to astronomy.net to obtain the WCS data


Jedijoe100:
There are a couple of problems with this, firstly not sure what to output. should this be a astropy WCS object? or should this be the raw header that is obtained from astroquery.

Secondly, there may be a number of errors the arise, and the question is how should these be handled? Also if astroquery returns no solution how do we want to hand this?

Also I am not sure where to place this function in the code.

trail/upload/process_uploads/header_standardizer.py
```diff
@@ -10,14 +10,18 @@
 from astropy.io.fits import PrimaryHDU, CompImageHDU
 from astropy.wcs import WCS
 import astropy.units as u
+from astroquery.astrometry_net import AstrometryNet
 
 import upload.models as models
 
 __all__ = ["HeaderStandardizer", ]
 
-
```
DinoBektesevic, line 18: Any reasons to remove this?

trail/upload/process_uploads/header_standardizer.py
```diff
@@ -10,14 +10,18 @@
...
 logger = logging.getLogger(__name__)
 
+ast = AstrometryNet()
+
+ast.api_key = 'ethexxgyljfuayob'  # currently the key connected to astronomy.net with the git account JediJoe100
```
DinoBektesevic: This is a bad idea, you should never commit private API keys to a public repository. Change this key as soon as you can. 
If special authentications are required they should be handled by the `SecretsConfig` class like they are in [settings.py](https://github.com/dirac-institute/trailblazer/blob/astrometry.net_wcs_solver/trail/trail/settings.py#L38). 

Given how changing file permissions and passing POSIX file-permissions tests is near impossible task on Windows it would seem, which is how we got the annoying `Config` and `SecretConfig` classes, I'd suggest you use a regular `Config` here, add another YAML file to site config dir and read the code from there.         
You must protect against the case when the file doesn't exists because we want this to work even if someone doesn't have the API key.

trail/upload/process_uploads/header_standardizer.py
```diff
@@ -98,7 +102,7 @@ def _computeStandardizedWcs(header, dimX, dimY):
         location. Corner is taken to be the (0,0)-th pixel.
         """
         standardizedWcs = {}
-        centerX, centerY = int(dimX/2), int(dimY/2)
+        centerX, centerY = int(dimX / 2), int(dimY / 2)
```
DinoBektesevic: If you're having flake8 errors make sure you're reading the same flake8 configuration (`trailblazer/setup.cfg`) that is used repo-wide.

trail/upload/process_uploads/header_standardizer.py
```diff
@@ -323,3 +328,36 @@ def standardizeWcs(self, hdu=None):
 
         wcs = models.Wcs(**self._computeStandardizedWcs(header, dimX, dimY))
         return wcs
+
+    def getWcs(self, path_to_file):
```
DinoBektesevic: Nothing calls this function?

`HeaderStandardizer`s have two functions: 1) standardize observation metadata (observer, location, time...) and 2) standardize image metadata (wcs only at the moment). At the moment standardizing WCS consists of attempting to read it with Astropy, and failing if it doesn't exist. 

What I wanted you to do is to to try to read it with Astropy, like it is now, and if that failed, solve it with astrometry.net solver. So the current content of `standardizeWcs` should be evacuated to the `_conmputeStandardizedHeader` and renamed to something like, for example, `_astropyWcsReader`. The `standardizeWcs` should be something like `if astropyWcsReader failed attempt astrometryNetSolver`

trail/upload/process_uploads/header_standardizer.py
```diff
@@ -10,14 +10,18 @@
...
+from astroquery.astrometry_net import AstrometryNet
 
 import upload.models as models
 
 __all__ = ["HeaderStandardizer", ]
 
-
 logger = logging.getLogger(__name__)
 
+ast = AstrometryNet()
```
DinoBektesevic: Can we name this something more explicit like `astronet_client`? I guess if it's going to be a global variable maybe we should name it with all caps too.

trail/upload/process_uploads/header_standardizer.py
```diff
@@ -323,3 +328,36 @@ def standardizeWcs(self, hdu=None):
...
+
+        Notes
+        -----
+        Currently this just ignores errors in the process.
+        It might be better to just raise the errors and let whatever the other function is to handle it.
+        """
+        try:
+            header = ast.solve_from_image(path_to_file, False)
+            # return that there was an error
+        except TimeoutError:
```
DinoBektesevic: Don't silence errors here. Let all the errors propagate all the way up to [`FitsProcessor.process`](https://github.com/dirac-institute/trailblazer/blob/main/trail/upload/process_uploads/fits_processor.py#L295) where we should, eventually, try to handle them or at the very lease propagate back to the user in the [`views.py`](https://github.com/dirac-institute/trailblazer/blob/main/trail/upload/views.py).

At best what we could do here is to define and raise a new exception like `class StandardizeWcsException(Exception)` and then `raise StandardizeWcsException(msg, attrs) from err`. 

What you should also do is add a `Raises` section to the docstring, instead of the `Notes` section, documenting what types of errors the function might raise.

requirements.txt
```diff
@@ -15,6 +16,7 @@ MarkupSafe==1.1.1
 matplotlib==3.3.4
 moto==2.0.5
 numpy==1.20.1
+photutils==1.1.0
```
DinoBektesevic, line 19: Did this come as a requirement of astroquery or is it a direct requirement for trailblazer? In any case I'm not sure where we use it.

trail/upload/process_uploads/header_standardizer.py
```diff
@@ -323,3 +328,36 @@ def standardizeWcs(self, hdu=None):
...
+            header = {}
+            # return that there was a timeout error
+        except TypeError:
+            header = {}
+            # there was an error in the fits file
+        return header  # return what needs to be parsed
```
DinoBektesevic: What is the return type here? It should be the `WCS` object from models.

DinoBektesevic (changes requested):
There's some work that still needs to be done but this is a really good start. Tests for astrometry.net function would be nice - I suspect it will be quite challenging to mock them.

requirements.txt
Jedijoe100: This is not referenced however, it speeds up the solve_from_image function. Basically not essential but speeds the process up.

DinoBektesevic: Ok, thanks for clarifying.

trail/upload/process_uploads/standardizers/moa_standardizer.py
```diff
@@ -69,16 +69,11 @@ def standardizeMetadata(self):
...
-            wcs_center_z=-999.999,
-            wcs_corner_x=-999.999,
-            wcs_corner_y=-999.999,
-            wcs_corner_z=-999.999
-        )
+        # ignores any WCS information in the header and instead just sends it to astrometry.net to solve.
+        # need to add a catch that catches the errors in astrometryNetSolver
+        try:
+            header, dimX, dimY = self._astrometryNetSolver(self._kwargs['filepath'])
+            wcs = Wcs(**self._computeStandardizedWcs(header, dimX, dimY))
```
mrawls: Suggest having just one line in `try` block if feasible. Pretty sure you're catching the exception on the first line; assuming that's true, I suggest moving the second line to an `else` block (following `except`).

trail/upload/process_uploads/header_standardizer.py
```diff
@@ -320,6 +379,50 @@ def standardizeWcs(self, hdu=None):
...
+        RuntimeError
+            There is no astrometry.net key
+        TimeoutError
+            If it does not find a solution with the given time. Does not mean
+            that astrometry.net cannot solve it.
+        TypeError
+            Some error in fits file preventing astrometry.net from working
+        FileNotFoundError
+            File not found from the path to file
+
```
mrawls, line 415: Please add `Returns` docstring block

trail/upload/process_uploads/header_standardizer.py
```diff
@@ -280,6 +301,44 @@ def standardizeWcs(self, hdu=None):
...
+        ----------
+        hdu : `object` or `None`, optional
+            An Astropy image-like HDU unit. Useful when dealing with
+            mutli-extension fits files where metadata is in the PrimaryHDU but
+            the WCS and image data are stored in the extensions.
+
+        Returns
+            -------
+            standardizedWCS : `upload.models.Wcs`
+            Standardized WCS keys and values.
```
mrawls:
```suggestion
        -------
        standardizedWCS : `upload.models.Wcs`
        Standardized WCS keys and values.
```

mrawls: Maybe my suggestion isn't quite right either 😆 you get the idea, please fix the indentations

trail/upload/process_uploads/header_standardizer.py
```diff
@@ -320,6 +379,50 @@ def standardizeWcs(self, hdu=None):
...
+        return header, dimX, dimY
 
-        wcs = models.Wcs(**self._computeStandardizedWcs(header, dimX, dimY))
-        return wcs
+    @staticmethod
+    def _astrometryNetSolver(path_to_file):
+        """Given a fits file it will process and send to astrometry.net
+        where it will obtain the WCS data for the file if it is able to.
+        If not then it will return an empty dictionary
+        (this can be altered to whatever needs to be used for)
```
mrawls: Is this correct? It looks like it will either raise or return 3 things (header dict, dimX, dimY). Please update the Returns block and description to be self consistent.

trail/upload/tests/test_processors.py
```diff
@@ -10,11 +10,24 @@
 from upload.process_uploads.upload_wrapper import TemporaryUploadedFileWrapper
 from upload.process_uploads.upload_processor import UploadProcessor
 from upload.process_uploads.fits_processor import FitsProcessor
+import upload.process_uploads.header_standardizer as header_standardizer
 
 
 TESTDIR = os.path.abspath(os.path.dirname(__file__))
 
 
+class MockAstrometryServer:
```
mrawls, line 19: Please add a brief class docstring explaining what this is for, along with the `TestAstronometryNet` class below.

trail/upload/tests/test_processors.py
```diff
@@ -29,6 +42,26 @@ def temporary_file_path(self):
         return self.sourceFilePath
 
 
+class TestAstrometryNet(TestCase):
+    testDataDir = os.path.join(TESTDIR, "data")
+
+    def setup(self):
+        pass
+
+    def testNoSolution(self):
```
mrawls, line 54: Again, one-liner docstrings please, stating each test is supposed to raise whatever error in whatever situation.

mrawls (commented):
This is looking great! All my comments are minor, so unless Dino sees anything else that needs tidying up, I am very happy 😄 Thanks in particular for renaming files to "cutout" for clarity.

If you are up for it, bonus points if you fixup/squash your commits into 3-4 total commits to make the history a bit neater.

trail/upload/process_uploads/standardizers/moa_standardizer.py
Jedijoe100: Do you mean in the `_astrometryNetSolver` function? Because that is not the function that has the `try` block it only raises errors.

trail/upload/process_uploads/standardizers/moa_standardizer.py
```diff
@@ -73,7 +73,6 @@ def standardizeWcs(self, **kwargs):
         # need to add a catch that catches the errors in astrometryNetSolver
         try:
             header, dimX, dimY = self._astrometryNetSolver(self._kwargs['filepath'])
-            wcs = Wcs(**self._computeStandardizedWcs(header, dimX, dimY))
         except ValueError and RuntimeError and TypeError as err:
             raise StandardizeWcsException("Failed to standardize WCS") from err
-        return wcs
+        return Wcs(**self._computeStandardizedWcs(header, dimX, dimY))
```
mrawls, line 78: Good solution 😀

trail/upload/process_uploads/header_standardizer.py
```diff
@@ -268,6 +289,44 @@ def standardizeMetadata(self):
...
+        Notes
+        -----
+        Tries to read the WCS from the header.
+        if that does not work then gives it to astrometry.net to look for a solution for the WCS
+        """
+        try:
+            astropy_fail = False
+            try:
+                header, dimX, dimY = self._astropyWcsReader(hdu)
+            except ValueError and TypeError:
```
DinoBektesevic: Shorten this to:

```
except (ValueError, TypeError):
    header, dimX, dimY = self._astrometryNetSolver(self.filepath)
````

no `if`s required I believe. The parenthesis is to conform to the [python manual](https://docs.python.org/3/tutorial/errors.html#handling-exceptions):
> An except clause may name multiple exceptions as a parenthesized tuple, for example:

File: trail/upload/process_uploads/header_standardizer.py [hunk 17]
```diff
@@ -268,6 +289,44 @@ def standardizeMetadata(self):
...
+                astropy_fail = True
+            if astropy_fail:
+                header, dimX, dimY = self._astrometryNetSolver(self.filepath)
+            wcs = models.Wcs(**self._computeStandardizedWcs(header, dimX, dimY))
```
DinoBektesevic: Move this line to the last line above the return, so that it's outside of the try-except blocks. This is mostly stylistic - saying that we except certain processing errors, but failure to instantiate is the problem of the `Wcs` class, not processing.

trail/upload/tests/test_processors.py
```diff
@@ -10,11 +10,25 @@
...
 
 TESTDIR = os.path.abspath(os.path.dirname(__file__))
 
 
+class MockAstrometryServer:
+    """Mock Astrometry function that acts like the astrometry.net server"""
+    def __init__(self, return_no_solution=False):
+        self.return_no_solution = return_no_solution
+
+    def solve_from_image(self, path_to_file, preprocess=True, solve_timeout=120):
```
DinoBektesevic, line 24: Might consider adding a comment that `preprocess` is here to just match the call signature in `header_standardizer.py`, optionally to make this future-proof remove `preprocess` and add `**kwargs`. In that way it will never matter what the actual full signature of the mocked function could ever be - just what are key arguments you need in your tests.

trail/upload/process_uploads/header_standardizer.py
```diff
@@ -268,6 +289,44 @@ def standardizeMetadata(self):
...
+        except ValueError and RuntimeError and TypeError as err:
```
DinoBektesevic: Can you make this a tuple - I don't want to risk some weird behavior with `and`s.

DinoBektesevic (approved):
Minor stylistic comments, approved for merge with a note to myself to double check why Astropy returns a default WCS for MOA headers. I have a sneaky suspicion this happens for all headers with no WCS information and that we should treat it as an error in `header_standardizer.py:L318`.


Thanks for renaming test data files. This is great work!

PR #38, opened 2021-10-21
Pagination adjustments

Jedijoe100:
This pull request is some adjustments to the pagination, these are the following:

- making it so that not all the links to different pages are viewable to stop like 100 buttons being visible
- centred the pagination for views
- added a siteConfig value for the number of images viewed on the gallery page as previously was basically hardcoded.


mrawls (approved):
Belated thank you for your work, Joe!

PR #39, opened 2021-12-22
Upload app upload form update

DinoBektesevic:
Update upload app's upload form and have some fun with alliteration on a public platform. 

Added drag and drop, upload progress bar, ability to add and remove files, added user message for post upload waiting period (pending some better solution using celery or something) and added post-upload processing report. 

TODO: make it neater and add failure handling.


trail/upload/models.py
```diff
@@ -481,14 +482,127 @@ def toDict(self):
...
+class StandardizedResult:
+    """A dataclass that associates standardized header metadata with one or
+    more standardized WCS and their thumbnails.
+    """
+    _thumbkeys = ["thumbnails", "thumbs", "thumbnail", "thumb"]
+    header: StandardizedHeader = None
+    thumbnails: Sequence[Thumbnails] = field(default_factory=list)
+
+    @classmethod
+    def __resolve_thumbs(cls, data):
```
mrawls, line 496: Can any of this thumbnail handling be shared with the gallery app? Or too different of a situation?

trail/upload/models.py
```diff
@@ -481,14 +482,127 @@ def toDict(self):
...
+
+    @property
+    def metadata(self):
+        """Standardized metadata."""
+        return self.header.metadata
+
+    @property
+    def isMultiExt(self):
+        """True when the header is a multi extension header."""
+        return len(self.wcs) > 1
```
mrawls, line 586: Are there any situations where there could be more than one WCS but it's not actually multiple image extensions?

trail/upload/models.py
```diff
@@ -481,14 +482,127 @@ def toDict(self):
...
+
+    def appendThumbnail(self, thumbnail):
```
mrawls, line 588: Docstring

trail/upload/models.py
```diff
@@ -481,14 +482,127 @@ def toDict(self):
...
+        thumb = dataToComponent(thumbnail, Thumbnails)
+        if thumb is not None:
+            self.thumbnails.append(thumb)
+        else:
+            raise ValueError("Could not create a Thumbnail from the given data!")
+
+    def extendThumbnails(self, thumbnails):
```
mrawls, line 602: Docstring

trail/upload/models.py
```diff
@@ -481,14 +482,127 @@ def toDict(self):
...
+            self.appendThumbnail(thumb)
+
+    def toDict(self):
+        """Returns a dictionary of standardized metadata, wcs an thumbnails."""
+        metadataDict = self.header.toDict()
+        if self.isMultiExt:
+            thumbDicts = {"thumbnails": [thumb.toDict() for thumb in self.thumbnails]}
+        else:
+            thumbDicts = {"thumbnails": {"large": self.thumbnails[0].large,
+                                         "small": self.thumbnails[0].small}}
```
mrawls, line 619: Not clear to me why single-extension headers get large/small thumbnails but multi-extension ones don't?

trail/upload/process_uploads/fits_processor.py
```diff
@@ -299,33 +300,46 @@ def process(self):
...
+
+        standardizedResult.metadata.upload_info = self.uploadInfo
+        standardizedResult.metadata.save()
+        for wcs in standardizedResult.wcs:
+            wcs.metadata = standardizedResult.metadata
+            wcs.save()
 
         # Create thumbnails (their DB models and the files) and then set up
         # relationship between particular wcs data and thumbs; then insert them
         # TODO: I'm iffed how this is set here, maybe refactor?
```
mrawls, line 326: Did this get fixed by chance? 😆

trail/upload/process_uploads/fits_processor.py
```diff
@@ -299,33 +300,46 @@ def process(self):
...
         logger.info(f"ID {self.uploadInfo.id}: Creating thumbnails.")
         thumbnails = self.createThumbnails()
         if isinstance(thumbnails, Thumbnails):
-            for wcs in standardizedHeader.wcs:
-                thumbnails.wcs = wcs
-            thumbnails.save()
+            standardizedResult.appendThumbnail(thumbnails)
```
mrawls: A bit confusing since you appear to define the append method as working on a single thumbnail and the extend method as working on many thumbnails, but here you have both working on `thumbnails`...?

trail/upload/templates/upload_form.html
```diff
@@ -0,0 +1,100 @@
...
+                    <p class="help">{{ field.help_text|safe }}</p>
+                    {% endif %}
+                    {{ field.errors }}
+                </div>
+            {% endfor %}
+        </div>
+        {% endif %}
+
+        <div class="file_list_input">
+            <p>Select multiple files with the file dialog or by dragging and dropping them onto the dashed region</p>
```
mrawls: Drag and drop FITS files here, or click to select multiple files.
By uploading to Trailblazer, you affirm you have all necessary rights and permission to share the files, and agree to make all uploaded materials publicly accessible via Trailblazer under CC BY 4.0 (add a link to https://creativecommons.org/licenses/by/4.0).

We should probably have language even better than this, and/or a required checkbox, but this is a start.

trail/upload/templates/upload_form.html
```diff
@@ -0,0 +1,100 @@
...
+                     role="progressbar" aria-valuenow="0" aria-valuemin="0" aria-valuemax="100" style="width:0%">
+                </div>
+            </div>
+            <div class="uploadform_user_message_wrapper" style="padding:10px;padding-top:30px;padding-bottom:30px">
+                <div class="uploadform_user_message">
+                    You files have been uploaded! <br>
+                    We are applying oils and incantations. This could take a while. <br>
+                    Success report may or may not be displayed at the end of it. <br>
+                    It is safe to leave this page now. <br>
+                    Thank you.
```
mrawls: Your files are being uploaded!
This could take a bit. It is safe to leave this page now.
Thank you for contributing to Trailblazer!

mrawls: I find the bit about "success report may or may not be displayed" very confusing as a user haha

mrawls (approved):
Thanks for doing this! My only ask would be to ensure the upload page is as accessible as possible. I have limited experience here, but I think this mostly means ensuring there are good labels for a screen reader and that the buttons can be pressed with keyboard-only input. This site covers the basics: https://www.w3.org/WAI/tips/developing/

trail/upload/models.py
DinoBektesevic: Not on the front page of the gallery, but perhaps on the "single image" view. I could see what it would take to unify the code and use `StandardizedResult`. 

Overall, however, gallery is really all about displaying the thumbs and doesn't care much about metadata or wcs and this class is all about tying the two together so we could, but we wouldn't get that much use out of it yet.

trail/upload/models.py
DinoBektesevic: I've been thinking about that myself. Wcs are really the "property" of an detector and kind of describes the mapping between what that detector sees and the real life geometry. The only thing we are interested in from a detector is the image. So each image would have a WCS. Or at least that was my logic. 

That means that it's not possible to have an image of the sky and not have a Wcs, or if we do - we don't want that image because we don't know what we're looking at. It is possible, however, to have multiple images but only 1 Wcs -> i.e. mask planes etc...

This here however, was copied over from `fits_processor`, where the above is all applied and determined, but this here has nothing to do with that. It just checks if there are multiple Wcs's in the `self.wcs` list. It's just for simplicity really, I thought it neater to ask this than to ask for `if len(stdres.wcs)==1` when using it. Will fit docstring.

trail/upload/process_uploads/fits_processor.py
DinoBektesevic: Not really, this is the same thing as before, just in one place now. It's all the inheritance thing - until metadata is inserted you don't know its ID, so you can't make a complete Wcs ready for insertion, without inserting Wcs object you don't have its ID so you can't make it's Thumbnail object. Why bulk-inserting them doesn't work though - no idea. 

Previously a part of this was done in `StandardizedHeader` class, perhaps it's neater if we hide it there but I again can't seem to figure out how to correctly use bulk inserts. I hoped that moving it here would help, but it didn't. TODO should stay.

trail/upload/process_uploads/fits_processor.py
DinoBektesevic: `FitsProcessor` is a parent class implementing some generic functionality but punting much of it to the actual processors. Some of them (f.e. SDSS) will return a thumbnail a file, others several (f.e DECam). I can't know in advance. Perhaps better naming scheme? Perhaps I should standardize on returning a list with 1 element all the time and use extend everytime?

In any case, somewhere the list vs a single will have to be handled.

mrawls: Slightly more standardized names for these methods would get my vote for now, if it's not too much trouble. It's a bit tricky to follow what is happening with the current naming. (I think of appending and extending in terms of python list operations, of course with different results.) I get that handling images with multiple, uh, images in them is annoying.

DinoBektesevic: As a reminder: some processors return a single Thumbnail, some return a collection of them. 
When they return a single `Thumbnail` I `append` - when they return a collection, I `extend`. The mentioned problem was that I named the variable in which I store the thumbnail making-results into as `thumbnails` which was confusing. 

I hadn't addressed this (forgot), happy to hear any suggestions you might have.

mrawls: I thought it was the naming of append vs extend which was confusing me, but maybe it was really `thumbnails` all along. Maybe call it a thumbnail-group or something? So long as you disambiguate between "single thing that will render nicely on a webpage" vs "giant pile of every fits extension's thumbnail," I am happy.

PR #44, opened 2022-03-28
Swagger doc

didntpay:
Added swagger package to demo the documentation tools for REST API, the path is /query/demo

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -30,3 +30,4 @@ six==1.15.0
 SQLAlchemy==1.4.1
 sqlparse==0.4.1
 Werkzeug==1.0.1
+drf-yasg
--- a/trail/query/urls.py
+++ b/trail/query/urls.py
@@ -3,7 +3,7 @@ from django.urls import path, include
 
 from rest_framework import routers
 from rest_framework.urlpatterns import format_suffix_patterns
-
+from query import views
 """Route query URLs to query views here."""
 
 
@@ -15,6 +15,7 @@ urlpatterns = [
     path('', index, name='query'),
     path('results', print_results, name='results'),
     path('rest/', include(router.urls)),
+    path('demo/', views.MetaDataQuery.as_view()),
     path('api-auth/', include('rest_framework.urls', namespace='rest_framework'))
 ]
 
--- a/trail/query/urls.py
+++ b/trail/query/urls.py
@@ -1,11 +1,29 @@
-from query.views import print_results, index
-from django.urls import path
+from query.views import print_results, index, MetadataViewSet, get_metadata
+from django.urls import path, include
 
+from rest_framework import routers
+from rest_framework.urlpatterns import format_suffix_patterns
 
 """Route query URLs to query views here."""
 
 
+router = routers.DefaultRouter()
+router.register(r'metadata', MetadataViewSet)
+
+
 urlpatterns = [
-   path('', index, name='query'),
-   path('results', print_results, name='results'),
+    path('', index, name='query'),
+    path('results', print_results, name='results'),
+    path('rest/', include(router.urls)),
+    path('api-auth/', include('rest_framework.urls', namespace='rest_framework'))
 ]
+
+
+rest2_patterns = format_suffix_patterns([
+    path('rest2/metadata/INSTR=<str:instrument>&TEL=<str:telescope>&FIL=<str:filter_name>', get_metadata),
+])
+
+urlpatterns.extend(rest2_patterns)
+
+
+
--- a/trail/query/views.py
+++ b/trail/query/views.py
@@ -1,8 +1,10 @@
 from django import forms
 from django.apps import apps
 from django.shortcuts import render
-
-
+from rest_framework import status
+from rest_framework.views import APIView
+from drf_yasg.utils import swagger_auto_schema
+from drf_yasg import openapi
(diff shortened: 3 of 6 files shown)


requirements.txt
```diff
@@ -30,3 +30,4 @@ six==1.15.0
 SQLAlchemy==1.4.1
 sqlparse==0.4.1
 Werkzeug==1.0.1
+drf-yasg
```
DinoBektesevic, line 33: Set a particular frozen version of `drf-yasg` that you guarantee that it works. Leaving it hanging like this means yasg can update and break something. Checking out the branch and installing it gave me `drf-yasg==1.20.0` You can check what you have by doing `pip freeze`

trail/query/urls.py
```diff
@@ -1,11 +1,30 @@
-from query.views import print_results, index
-from django.urls import path
-
+from query.views import print_results, index, MetadataViewSet, get_metadata
```
DinoBektesevic, line 1: `MetadataViewSet` doesn't exist in `query.views` but something called `MetaDataQuery` does. CamelCase is usually best used to separate words, so `MetadataQuery` would be the capitalization.

trail/query/urls.py
```diff
@@ -1,11 +1,30 @@
...
+from django.urls import path, include
 
+from rest_framework import routers
+from rest_framework.urlpatterns import format_suffix_patterns
+from query import views
 """Route query URLs to query views here."""
 
 
+router = routers.DefaultRouter()
+router.register(r'metadata', MetadataViewSet)
```
DinoBektesevic, line 11: See above.

trail/query/urls.py
```diff
@@ -1,11 +1,30 @@
...
+
+
 urlpatterns = [
-   path('', index, name='query'),
-   path('results', print_results, name='results'),
+    path('', index, name='query'),
+    path('results', print_results, name='results'),
+    path('rest/', include(router.urls)),
+    path('demo/', views.MetaDataQuery.as_view()),
```
DinoBektesevic, line 18: This is never imported, see above.

trail/query/views.py
```diff
@@ -61,3 +63,16 @@ def print_results(request):
 
     return render(request, "query.html",
                   {"data": query_results, "wcsdata": wcs_list, "queryform": form, "render_table": True})
+
+@api_view(['GET'])
```
DinoBektesevic, line 67: This fails to import because the `api_view` decorator comes from `from rest_framework.decorators import api_view` which we did not import on top. Either something got broken in rebase or you accidentally overwrote the import.

trail/trail/urls.py
```diff
@@ -18,6 +18,22 @@
 from django.conf import settings
 from django.urls import path, re_path, include
 from django.views.generic import TemplateView
+from drf_yasg.views import get_schema_view
+from drf_yasg import openapi
+from rest_framework import permissions
+
+schema_view = get_schema_view(
+   openapi.Info(
+      title="TrailBlazer API",
```
DinoBektesevic, line 27: I feel like these keyword args would best be punted to the Config class which would read them from a file. See [this](https://github.com/dirac-institute/trailblazer/blob/main/trail/trail/settings.py#L31) and [this](https://github.com/dirac-institute/trailblazer/blob/main/trail/trail/config.py) for the Config class and [this](https://github.com/dirac-institute/trailblazer/tree/main/trail/config) for example of the files. 

None of these values are secrets so you can add them directly to the config directory. 

License information is in our repo, see [here](https://github.com/dirac-institute/trailblazer/blob/main/LICENSE)

Description could be something better than "Test PR" such as "Rest API Documentation", make sure we change this can stay like that but we should put an admin email there as soon as we manage.

trail/trail/urls.py
```diff
@@ -18,6 +18,22 @@
...
+      default_version='v1',
```
DinoBektesevic, line 28: We are giving a "default" version here - but where do we register what version is a particular API of? Or am I misunderstanding what this means (didn't read the docs).

trail/trail/urls.py
```diff
@@ -29,6 +45,7 @@
     # in the new template the index is the gallery, we don't need so many apps
     re_path(r'^index/', include('gallery.urls')),
     re_path(r'^',  include('gallery.urls')),
+    path('swagger', schema_view.with_ui(cache_timeout=0), name="swagger-json"),
```
DinoBektesevic, line 48: Probably better to name the link something like "api_docs" and add a link somewhere on the query page (open issue or trello ticket).

DinoBektesevic (changes requested):
There seems to be a problem with how you rebased the branch. As it is the branch does not run for me. Fixing it I find there are quite a bit of flake8 errors too. 

Rebase to the latest to include the latest PR as well a then test it runs locally.

The major comment I have, except make sure it actually works, is that I don't believe we have to have any kind of REST API actually implemented to make swagger work. The documentation will just be empty, but that's ok as it will update as we progress with that issue.           
Personally I would remove all mentions of any REST API from the urls, serializers etc. completely from the code; we just don't need it in `main` for swagger to still work correctly.

Thanks for the hard work!

DinoBektesevic:
Moved to PR-https://github.com/dirac-institute/trailblazer/pull/46

PR #45, opened 2022-03-31
Fix wrong path to upload_button_state JS script.

DinoBektesevic:
Path to JS script that controlled the upload button state was hardcoded in the HTML template with a wrong path. This accidentally worked locally but does not work in general. 

Path is now set in the `UploadForm` class itself and points to correct location.


DinoBektesevic:
[Fix wrong upload button link.](https://trello.com/c/bZiHnuQM/46-fix-wrong-upload-button-link)

arihani2 (approved):

PR #46, opened 2022-04-19
Query restapi

didntpay:
Implemented Metadata table api and added swagger doc support. The post parameter are not yet listed on the swagger doc, that is expected since we need to manually add to it.

Also implemented basic functional test for metadata query.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,4 +7,5 @@ moto==3.0.3
 numpy==1.22.2
 Pillow==9.0.1
 pyqt5>=5.0
+djangorestframework==3.13.1
 drf-yasg==1.20.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,3 +7,4 @@ moto==3.0.3
 numpy==1.22.2
 Pillow==9.0.1
 pyqt5>=5.0
+drf-yasg==1.20.0
--- a/trail/config/site.yaml
+++ b/trail/config/site.yaml
@@ -7,3 +7,12 @@ astrometry_net:
   key:
   timeout: 200
 gallery_image_count: 12
+api_doc:
+  title: "TrailBlazer API"
+  version: "v1"
+  description: "Add description later!"
+  contact_email: "changethis@gmail.com"
+  license:
+    name: "MIT License"
+    url: "https://github.com/dirac-institute/trailblazer/blob/main/LICENSE"
+    
--- a/trail/config/site.yaml
+++ b/trail/config/site.yaml
@@ -10,7 +10,7 @@ gallery_image_count: 12
 api_doc:
   title: "TrailBlazer API"
   version: "v1"
-  description: "Add description later!"
+  description: "Trailblazer backend rest API"
   contact_email: "changethis@gmail.com"
   license:
     name: "MIT License"
--- a/trail/config/site.yaml
+++ b/trail/config/site.yaml
@@ -8,7 +8,7 @@ astrometry_net:
   timeout: 200
 gallery_image_count: 12
 api_doc:
-  title: "TrailBlazer API"
+  title: "Trailblazer API"
   version: "v1"
   description: "Trailblazer backend rest API"
   contact_email: "changethis@gmail.com"
--- a/trail/query/tests.py
+++ b/trail/query/tests.py
@@ -1,16 +1,12 @@
 from django.test import TestCase
 from upload.models import Metadata
-from upload.serializers import MetadataSerializer
-from rest_framework.test import APIClient
-from upload.models import UploadInfo
-from rest_framework import status
-import io
-from rest_framework.parsers import JSONParser
+from upload.models import UploadInfo, Wcs
+from query.views import MetadataDAO
 # Create your tests here.
 
 
 class MetadataQueryTest(TestCase):
-    client = APIClient()
+    metadataDao = MetadataDAO()
 
     def setUp(self):
         info = UploadInfo()
@@ -46,18 +42,34 @@ class MetadataQueryTest(TestCase):
             exposure_duration=30.0,
             filter_name='i'
         )
+        wcs1 = {
+            'metadata': meta,
+            'wcs_radius': 1,
+            'wcs_center_x': 0.4,
+            'wcs_center_y': 0.3,
+            'wcs_center_z': 0.7,
+            'wcs_corner_x': 16,
+            'wcs_corner_y': 17,
+            'wcs_corner_z': 0.7
+        }
         meta.save()
         meta2.save()
 
+        self.wcs = Wcs(**wcs1)
+        self.wcs.save()
+
     def testBasicQueries(self):
-        response = self.client.get("/query/getMetadata?processor_name=fits")
-        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        """Tests getMetadataByParams in metadataDao is functional"""
+        queryParam = {"processor_name__icontains": "fits"}
+        response = self.metadataDao.getMetadataByParams(queryParam)
 
-        stream = io.BytesIO(response.content)
-        data = JSONParser().parse(stream)
+        self.assertEqual(len(response), 2)
+        for metadata in response:
+            self.assertTrue("fits" in metadata.processor_name.lower())
 
-        metadatas = MetadataSerializer(data=data, many=True)
+    def testBasicWcsQuery(self):
+        """Tests getMeatadataInSpecifiedSky in metadataDao is functional"""
+        queryParam = {"raLow": 0, "raHigh": 1000, "decLow": 0, "decHigh": 1000}
+        response = self.metadataDao.getMeatadataInSpecifiedSky(queryParam)
 
-        self.assertTrue(metadatas.is_valid(raise_exception=True))
-        result = metadatas.create(metadatas.validated_data)
-        self.assertEqual(len(result), 2)
+        self.assertEqual(len(response), 1)
--- a/trail/query/tests.py
+++ b/trail/query/tests.py
@@ -1,3 +1,79 @@
-# from django.test import TestCase
-
+from django.test import TestCase
+from upload.models import Metadata
+from django.test import Client
+from rest_framework.test import APIClient
+import json
+from upload.models import UploadInfo
 # Create your tests here.
+class MetadataQueryTest(TestCase):
+    client = APIClient()
+
+    def setUp(self):
+        info = UploadInfo()
+        info.save()
+
+        meta = Metadata(
+            upload_info=info,
+            processor_name="MultiExtensionFits",
+            standardizer_name="",
+            instrument="",
+            telescope="",
+            science_program="",
+            obs_lon=-101,
+            obs_lat=1.0,
+            obs_height=1.0,
+            datetime_begin='2009-10-17T01:53:42.1',
+            datetime_end='2009-10-18T01:53:42.1',
+            exposure_duration=30.0,
+            filter_name='i'
+        )
+        meta2 = Metadata(
+            upload_info=info,
+            processor_name="DECamCommunityFits",
+            standardizer_name="",
+            instrument="",
+            telescope="",
+            science_program="",
+            obs_lon=10,
+            obs_lat=1.0,
+            obs_height=1.0,
+            datetime_begin='2009-10-17T01:53:42.1',
+            datetime_end='2009-10-18T01:53:42.1',
+            exposure_duration=30.0,
+            filter_name='i'
+        )
+        meta.save()
+        meta2.save()
+    
+    def test_or_relationship_when_two_query_set(self):
+        query1 = {"processor_name" : "MultiExtensionFits"}
+        query2 = {"processor_name" : "DECamCommunityFits"}
+        requestReturnCols = ["obs_lon", "processor_name"]
+        requestPayload = self.createRequestPayload(returnCols=requestReturnCols, queryParams=[query1, query2])
+        reponse = self.client.post("/query/getMetadata", json.dumps(requestPayload), content_type='application/json')
+        self.assertEqual(reponse.status_code, 200)
+        jsonContents = json.loads(reponse.content)
+        for content in jsonContents:
+            if (content["processor_name"] != "MultiExtensionFits" and content["processor_name"] != "DECamCommunityFits"):
+                self.assertEqual(content["processor_name"], "")
+
+    def test_and_relationship_when_two_param_set(self):
+        query1 = {"processor_name" : "MultiExtensionFits", "obs_lon__gte" : "-100", "obs_lon__lte" : "100"}
+        requestReturnCols = ["obs_lon", "processor_name"]
+        requestPayload = self.createRequestPayload(returnCols=requestReturnCols, queryParams=[query1])
+        reponse = self.client.post("/query/getMetadata", json.dumps(requestPayload), content_type='application/json')
+        self.assertEqual(reponse.status_code, 200)
+        jsonContents = json.loads(reponse.content)
+        for content in jsonContents:
+            self.assertEqual(content["processor_name"], "MultiExtensionFits")
+            self.assertGreaterEqual(content["obs_lon"], -100)
+            self.assertLessEqual(content["obs_lon"], 100)
+    
+    def createRequestPayload(self, returnAllCols=1, returnCols=[], queryParams=[]):
+        request = {
+            "returnAllCols" : returnAllCols,
+            "returnCols" : returnCols,
+            "queryParams" : queryParams
+        }
+        return request
+        
--- a/trail/query/tests.py
+++ b/trail/query/tests.py
@@ -61,7 +61,7 @@ class MetadataQueryTest(TestCase):
     def testBasicQueries(self):
         """Tests getMetadataByParams in metadataDao is functional"""
         queryParam = {"processor_name__icontains": "fits"}
-        response = self.metadataDao.getMetadataByParams(queryParam)
+        response = self.metadataDao.queryByParams(queryParam)
 
         self.assertEqual(len(response), 2)
         for metadata in response:
@@ -70,6 +70,6 @@ class MetadataQueryTest(TestCase):
     def testBasicWcsQuery(self):
         """Tests getMeatadataInSpecifiedSky in metadataDao is functional"""
         queryParam = {"raLow": 0, "raHigh": 1000, "decLow": 0, "decHigh": 1000}
-        response = self.metadataDao.getMeatadataInSpecifiedSky(queryParam)
+        response = self.metadataDao.queryBySpecifiedSky(queryParam)
 
         self.assertEqual(len(response), 1)
--- a/trail/query/tests.py
+++ b/trail/query/tests.py
@@ -1,7 +1,6 @@
 from django.test import TestCase
 from upload.models import Metadata
 from rest_framework.test import APIClient
-import json
 from upload.models import UploadInfo
 # Create your tests here.
 
@@ -45,36 +44,3 @@ class MetadataQueryTest(TestCase):
         )
         meta.save()
         meta2.save()
-
-    def test_or_relationship_when_two_query_set(self):
-        query1 = {"processor_name": "MultiExtensionFits"}
-        query2 = {"processor_name": "DECamCommunityFits"}
-        requestReturnCols = ["obs_lon", "processor_name"]
-        requestPayload = self.createRequestPayload(returnCols=requestReturnCols, queryParams=[query1, query2])
-        reponse = self.client.post("/query/getMetadata", json.dumps(requestPayload), content_type='application/json')
-        self.assertEqual(reponse.status_code, 200)
-        jsonContents = json.loads(reponse.content)
-        for content in jsonContents:
-            if (content["processor_name"] != "MultiExtensionFits"
-                    and content["processor_name"] != "DECamCommunityFits"):
-                self.assertEqual(content["processor_name"], "")
-
-    def test_and_relationship_when_two_param_set(self):
-        query1 = {"processor_name": "MultiExtensionFits", "obs_lon__gte": "-100", "obs_lon__lte": "100"}
-        requestReturnCols = ["obs_lon", "processor_name"]
-        requestPayload = self.createRequestPayload(returnCols=requestReturnCols, queryParams=[query1])
-        reponse = self.client.post("/query/getMetadata", json.dumps(requestPayload), content_type='application/json')
-        self.assertEqual(reponse.status_code, 200)
-        jsonContents = json.loads(reponse.content)
-        for content in jsonContents:
-            self.assertEqual(content["processor_name"], "MultiExtensionFits")
-            self.assertGreaterEqual(content["obs_lon"], -100)
-            self.assertLessEqual(content["obs_lon"], 100)
-
-    def createRequestPayload(self, returnAllCols=1, returnCols=[], queryParams=[]):
-        request = {
-            "returnAllCols": returnAllCols,
-            "returnCols": returnCols,
-            "queryParams": queryParams
-        }
-        return request
--- a/trail/query/tests.py
+++ b/trail/query/tests.py
@@ -2,6 +2,9 @@ from django.test import TestCase
 from upload.models import Metadata
 from upload.models import UploadInfo, Wcs
 from query.views import MetadataDAO
+from coordinates import getXYZFromWcs, getRaDecFromXYZ
+import numpy as np
+
 # Create your tests here.
 
 
@@ -73,3 +76,17 @@ class MetadataQueryTest(TestCase):
         response = self.metadataDao.queryBySpecifiedSky(queryParam)
 
         self.assertEqual(len(response), 1)
+
+    def testCoordinateConversion(self):
+        ra = 200 * np.pi / 180
+        dec = 87 * np.pi / 180
+
+        cartersian = getXYZFromWcs(ra, dec)
+
+        self.assertTrue(abs(cartersian["x"] - (-0.049) < 0.1))
+        self.assertTrue(abs(cartersian["y"] - (-0.018) < 0.1))
+        self.assertTrue(abs(cartersian["z"] - (0.9986) < 0.1))
+
+        backToPolar = getRaDecFromXYZ(cartersian["x"], cartersian["y"], cartersian["z"])
+        self.assertTrue(abs(backToPolar["ra"] - ra < 0.1))
+        self.assertTrue(abs(backToPolar["dec"] - dec < 0.1))
--- a/trail/query/tests.py
+++ b/trail/query/tests.py
@@ -1,7 +1,11 @@
 from django.test import TestCase
 from upload.models import Metadata
+from upload.serializers import MetadataSerializer
 from rest_framework.test import APIClient
 from upload.models import UploadInfo
+from rest_framework import status
+import io
+from rest_framework.parsers import JSONParser
 # Create your tests here.
 
 
@@ -15,10 +19,10 @@ class MetadataQueryTest(TestCase):
         meta = Metadata(
             upload_info=info,
             processor_name="MultiExtensionFits",
-            standardizer_name="",
-            instrument="",
-            telescope="",
-            science_program="",
+            standardizer_name="astro_metadaor",
+            instrument="HSC",
+            telescope="Subaru",
+            science_program="asdfasdf",
             obs_lon=-101,
             obs_lat=1.0,
             obs_height=1.0,
@@ -29,12 +33,12 @@ class MetadataQueryTest(TestCase):
         )
         meta2 = Metadata(
             upload_info=info,
-            processor_name="DECamCommunityFits",
-            standardizer_name="",
-            instrument="",
-            telescope="",
-            science_program="",
-            obs_lon=10,
+            processor_name="SingleExtensionFits",
+            standardizer_name="astro_metadaor",
+            instrument="HSC",
+            telescope="Subaru",
+            science_program="asdfasdf",
+            obs_lon=-101,
             obs_lat=1.0,
             obs_height=1.0,
             datetime_begin='2009-10-17T01:53:42.1',
@@ -44,3 +48,16 @@ class MetadataQueryTest(TestCase):
         )
         meta.save()
         meta2.save()
+
+    def testBasicQueries(self):
+        response = self.client.get("/query/getMetadata?processor_name=fits")
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+
+        stream = io.BytesIO(response.content)
+        data = JSONParser().parse(stream)
+
+        metadatas = MetadataSerializer(data=data, many=True)
+
+        self.assertTrue(metadatas.is_valid(raise_exception=True))
+        result = metadatas.create(metadatas.validated_data)
+        self.assertEqual(len(result), 2)
--- a/trail/query/tests.py
+++ b/trail/query/tests.py
@@ -1,3 +1,52 @@
-# from django.test import TestCase
-
+from django.test import TestCase
+from upload.models import Metadata
+from django.test import Client
+from rest_framework.test import APIClient
+import json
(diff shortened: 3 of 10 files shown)


trail/config/site.yaml
```diff
@@ -7,3 +7,12 @@ astrometry_net:
   key:
   timeout: 200
 gallery_image_count: 12
+api_doc:
+  title: "TrailBlazer API"
+  version: "v1"
+  description: "Add description later!"
```
DinoBektesevic: Can we add a description now? Something like `Trailblazer metadata query API` would be good.

trail/query/tests.py
```diff
@@ -1,3 +1,80 @@
...
+            obs_height=1.0,
+            datetime_begin='2009-10-17T01:53:42.1',
+            datetime_end='2009-10-18T01:53:42.1',
+            exposure_duration=30.0,
+            filter_name='i'
+        )
+        meta.save()
+        meta2.save()
+
+    def test_or_relationship_when_two_query_set(self):
```
DinoBektesevic: If you add a 1 line short doc-strings to these tests they will print out when running the tests, instead they print less clear class name dot method name type output. Something like "Tests correct OR query execution" for example.

trail/query/urls.py
```diff
@@ -1,4 +1,4 @@
-from query.views import print_results, index
+from query.views import print_results, index, MetaDataQuery
```
DinoBektesevic: Can we rename the `MetaDataQuery` into `MetadataQuery`?

trail/query/views.py
```diff
@@ -61,3 +61,29 @@ def print_results(request):
 
     return render(request, "query.html",
                   {"data": query_results, "wcsdata": wcs_list, "queryform": form, "render_table": True})
+
+
+class MetaDataQuery(APIView):
```
DinoBektesevic: These classes and methods could really do with some documentation. I agree this isn't obvious throughout the whole codebase but parts of it do have pretty decent documentation, see [process uploads](https://github.com/dirac-institute/trailblazer/blob/main/trail/upload/process_uploads/header_standardizer.py) module for examples.

trail/query/views.py
```diff
@@ -61,3 +61,29 @@ def print_results(request):
...
+    DATA_RETURNCOLS = "returnCols"
+    DATA_QUERYPARAM = "queryParams"
+    DATA_RETURNALL = "returnAllCols"
+
+    def post(self, request):
```
DinoBektesevic: Missing docstring.

trail/query/views.py
```diff
@@ -61,3 +61,29 @@ def print_results(request):
...
+        returnAll = request.data[self.DATA_RETURNALL]
+        returnCols = request.data[self.DATA_RETURNCOLS]
+        queryParams = request.data[self.DATA_QUERYPARAM]
```
DinoBektesevic: What happens if some of these keys don't exist in the given request (hint: this should also be a test).

trail/query/tests.py
```diff
@@ -1,3 +1,80 @@
...
+        for content in jsonContents:
+            if (content["processor_name"] != "MultiExtensionFits"
+                    and content["processor_name"] != "DECamCommunityFits"):
+                self.assertEqual(content["processor_name"], "")
+
+    def test_and_relationship_when_two_param_set(self):
+        query1 = {"processor_name": "MultiExtensionFits", "obs_lon__gte": "-100", "obs_lon__lte": "100"}
+        requestReturnCols = ["obs_lon", "processor_name"]
+        requestPayload = self.createRequestPayload(returnCols=requestReturnCols, queryParams=[query1])
+        reponse = self.client.post("/query/getMetadata", json.dumps(requestPayload), content_type='application/json')
```
DinoBektesevic: Misspelled `response`

trail/query/views.py
```diff
@@ -61,3 +61,29 @@ def print_results(request):
...
+            # "queryParams" : [{"observer" : "me", "location" : "seattle"}]
+            iterationSet = querySet
+            if not (returnAll == 1):
+                iterationSet = iterationSet.filter(**queryParam).values(*returnCols)
+            else:
+                iterationSet = iterationSet.filter(**queryParam).values()
+            resultSet = resultSet | iterationSet
+
+        result = json.dumps(list(resultSet), default=str)
+        return HttpResponse(result, content_type="application/json")
```
DinoBektesevic: This whole thing could be simplified and made faster by just doing:

```
results = Metadata.objects.all()
for queryParam in queryParams:
    if returnAll:
        results = results.filter(**queryParam).values()
    else:
        results = results.filter(**queryParam).values(*returnCols)
```

trail/trail/settings.py
```diff
@@ -46,6 +45,13 @@
 DATA_ROOT = siteConfig.resolveAbsFromOrigin(siteConfig.data_root)
 
 
+API_DOC_TITLE = siteConfig.api_doc.title
+API_DOC_DEFAULT_VER = siteConfig.api_doc.version
+API_DOC_DESCRIPTION = siteConfig.api_doc.description
+API_DOC_CONTACT_EMAIL = siteConfig.api_doc.contact_email
+API_DOC_LICENSE_NAME = siteConfig.api_doc.license.name
+API_DOC_LICENSE_URL = siteConfig.api_doc.license.url
```
DinoBektesevic, line 53: Part of the existence of config class is that we don't have to do this mapping in here like this and keep the `settings.py` clean-er and more configurable. This feels like a thing we should add to the `toDict` method. Can you remind me who requires us to declare these variables? I would like to browse their documentation a bit.

If this is here just so we can declare if for `drf_yasg` below - then we can figure out a better way to deal with this.

trail/query/views.py
DinoBektesevic: Bonus points if you look for a `get_query` function at the top of the file and use it to immediately add "contains" and case sensitivity to the queries by pre-processing the keys you are querying on.

requirements.txt
```diff
@@ -7,3 +7,4 @@ moto==3.0.3
 numpy==1.22.2
 Pillow==9.0.1
 pyqt5>=5.0
+drf-yasg==1.20.0
```
DinoBektesevic, line 11: We should probably add the `djangorestframework` at some point.

trail/trail/settings.py
```diff
@@ -88,6 +94,7 @@
     'upload',
     'query',
     'trail',
+    'drf_yasg',
```
DinoBektesevic, line 98: See top-level comment.

DinoBektesevic (changes requested):
I am willing to accept this into the codebase, pending changes to comments in the code, because I think it's a really good start to getting the API implemented.

To make sure, however, that we are exactly on the same page - this isn't actually using `rest_framework`. What you did was you map a URL to a different view (class). You use that to return `HttpRequests` that contain data based on some data in the received request. This is the underlying principle of how a REST api should work -  true - but now you are stuck repeating
```
result = json.dumps(list(resultSet), default=str)
result = json.loads(reponse.content)
```
and hitting all the associated pitfalls that come with this.

Instead of doing that, leverage `rest_framework` functionality. At the moment, if you were to drop `rest_framework` imports completely and replaced removed the `APIView` with some built-in Django View class you wouldn't be able to tell the difference from functionality alone. 

I would have a look at the [`rest_framework` tutorial pages](https://www.django-rest-framework.org/tutorial/1-serialization/#creating-a-serializer-class) with more attention to what the tutorial is trying to show you. It will start slowly, by mapping a single table and individual columns, using `Serializers`. It will then show how to define a REST API by mapping urls to functions that return serializers. It will then discuss why sometimes relying on functions only is not the best way and introduce class based API. The point is to leverage this mapping between `Serializers` and `Models` and let them do all the work of serializing loads of DB data into JSON strings. 

Like I said, this can be merged, it's a good start and we can use a lot of the code that exists in this PR with the rest_framework.

trail/query/views.py
```diff
@@ -1,12 +1,13 @@
...
-
-Metadata = apps.get_model('upload', 'Metadata')
+from rest_framework.views import APIView
+from upload.models import Metadata
+from upload.models import Wcs
+import numpy as np
 
 
 class MetadataForm(forms.Form):
+
```
DinoBektesevic, line 14: Do we need this newline?

trail/query/views.py
```diff
@@ -61,3 +62,101 @@ def print_results(request):
...
+    DECLOW = "decLow"
+    DECHIGH = "decHigh"
+
+    """Support Query that returns metadata entry as a result through metadata info
+
+    Notes
+    -----
+    """
+    # gets all the metadata entry that matches the query parameters.
+    def getMetadataByParams(self, paramDict):
```
DinoBektesevic: Can we rename this into something shorter like `queryParams`? We are already dealing with `MetadataDAO` so it's obvious what we're querying on.

trail/query/views.py
```diff
@@ -61,3 +62,101 @@ def print_results(request):
...
+
```
DinoBektesevic, line 83: Remove empty line.

trail/query/views.py
```diff
@@ -61,3 +62,101 @@ def print_results(request):
...
+        queryParam: list of criterias for metadata
+
+        """
+
+        querySet = Metadata.objects.all()
+
+        if paramDict is not None:
+            querySet = querySet.filter(**paramDict)
+
+        return list(querySet)
```
DinoBektesevic, line 106: Is there a particular reason not to use serializers here? As-is it;s completely consistent with the documented return. Is this the "doesn't work when fields are specified" issue?

trail/query/views.py
```diff
@@ -61,3 +62,101 @@ def print_results(request):
...
+
+    def getMeatadataInSpecifiedSky(self, paramDict):
```
DinoBektesevic: Can we rename this into something shorter like `querySkyArea`? We are already dealing with `MetadataDAO` so it's obvious what we're querying on.

trail/query/views.py
```diff
@@ -61,3 +62,101 @@ def print_results(request):
...
+        Parameters
+        ----------------
+        raLow: the lower bound for ra
+        raHigh: the upper bound for ra
+        decLow: the lower bound for dec
+        decHigh: the upper bound for dec
+
+        Note
+        -------------------
+        if the parameters is not specified correctly, will return a bad request
```
DinoBektesevic: Lines 121 and 122 (couple of lines down) would disagree with this statement. This method seems to return an empty list when parameters are not specified correctly. It would also be good to describe what you mean by correctly, matches on the name only or matches on the name and additional factors too?

trail/query/views.py
```diff
@@ -61,3 +62,101 @@ def print_results(request):
...
+                                       float(paramDict.get(self.DECHIGH)))
+
+        filteredWcs = Wcs.objects.all().filter(**self.makeFilterDictForWcs(upperLeft, lowerRight))
+        # for every wcs, find the metadata id
+        metadataIds = set([metadataId['metadata'] for metadataId in filteredWcs.values('metadata')])
+        metadatas = self.getMetadatasByIds(metadataIds)
+
+        return metadatas
+
+    def getMetadatasByIds(self, metadataIds):
```
DinoBektesevic, line 149: A shorter name like `queryIds` perhaps?

trail/config/site.yaml
```diff
@@ -7,3 +7,12 @@ astrometry_net:
   key:
   timeout: 200
 gallery_image_count: 12
+api_doc:
+  title: "TrailBlazer API"
+  version: "v1"
+  description: "Trailblazer backend rest API"
+  contact_email: "changethis@gmail.com"
```
mrawls, line 14: @DinoBektesevic - any updates toward getting us a proper email?

trail/config/site.yaml
```diff
@@ -7,3 +7,12 @@ astrometry_net:
   key:
   timeout: 200
 gallery_image_count: 12
+api_doc:
+  title: "TrailBlazer API"
```
mrawls: Please just "Trailblazer", no capital B

trail/query/tests.py
```diff
@@ -1,3 +1,75 @@
-# from django.test import TestCase
-
+from django.test import TestCase
+from upload.models import Metadata
+from upload.models import UploadInfo, Wcs
+from query.views import MetadataDAO
 # Create your tests here.
+
+
+class MetadataQueryTest(TestCase):
```
mrawls, line 11: Please add a short class docstring (can be a one-liner) explaining the purpose

trail/query/views.py
```diff
@@ -61,3 +66,162 @@ def print_results(request):
...
+    Notes
+    -----
+    """
+
+    RALOW = "raLow"
+    RAHIGH = "raHigh"
+    DECLOW = "decLow"
+    DECHIGH = "decHigh"
+
+    """Support Query that returns metadata entry as a result through metadata info
```
mrawls: Please consolidate into a single class docstring at the top

trail/query/views.py
```diff
@@ -61,3 +66,162 @@ def print_results(request):
...
+
+    Notes
+    -----
+    """
+    # gets all the metadata entry that matches the query parameters.
+    def getMetadataByParams(self, paramDict):
+
+        """
+        Returns the Metadata entries that meets the users input
+        Parameters
```
mrawls, line 87: - Please add a blank line before Parameters
- You list `fields` and `queryParam` as Parameters, but I only see `paramDict`?
- Missing a Returns docstring section

trail/query/views.py
```diff
@@ -61,3 +66,162 @@ def print_results(request):
...
+
+        """
+
+        querySet = Metadata.objects.all()
+        if paramDict is not None:
+            querySet = querySet.filter(**paramDict)
+
+        return list(querySet)
+
+    def getMeatadataInSpecifiedSky(self, paramDict):
```
mrawls: Typo - "Metadata" not "Meatadata" 😄 be sure to also update each place this is called.

trail/query/views.py
```diff
@@ -61,3 +66,162 @@ def print_results(request):
...
+
+        """
+        Returns the Metadata associated with the part of sky the user specifies
+        ----------
+
+        Parameters
```
mrawls, line 113: The parameters listed here aren't inputs to this function, and should be documented elsewhere

trail/query/views.py
```diff
@@ -61,3 +66,162 @@ def print_results(request):
...
+        ----------------
+        raLow: the lower bound for ra
+        raHigh: the upper bound for ra
+        decLow: the lower bound for dec
+        decHigh: the upper bound for dec
+
+        Note
+        -------------------
+        if the parameters is not specified correctly, will return a bad request
```
mrawls: What kind of bad request? An error message that will help the user understand what happened and what they could do differently? This could be a good place to add a "Raises" section to the docstring, if a specific error is raised in this situation

trail/query/views.py
```diff
@@ -61,3 +66,162 @@ def print_results(request):
...
+                                       float(paramDict.get(self.DECHIGH)))
+
+        filteredWcs = Wcs.objects.all().filter(**self.makeFilterDictForWcs(upperLeft, lowerRight))
+        # for every wcs, find the metadata id
+        metadataIds = set([metadataId['metadata'] for metadataId in filteredWcs.values('metadata')])
+        metadatas = self.getMetadatasByIds(metadataIds)
+
+        return metadatas
+
+    def getMetadatasByIds(self, metadataIds):
```
mrawls, line 149: These and the few below are tiny functions, but I would still appreciate one-liner docstrings for each

trail/query/views.py
```diff
@@ -61,3 +66,162 @@ def print_results(request):
...
+
+
+class WcsQuery(APIView):
+
+    metadataDao = MetadataDAO()
+
+    def get(self, request):
+
+        """
+        Returns the Metadata associated with the part of sky the user specifies
```
mrawls, line 272: Please tidy up this getter docstring, and add a class docstring

trail/trail/settings.py
mrawls: Was this resolved?

trail/trail/urls.py
```diff
@@ -18,11 +18,31 @@
...
+from drf_yasg.views import get_schema_view
+from drf_yasg import openapi
+from rest_framework import permissions
+
+schema_view = get_schema_view(
+   openapi.Info(
+      title=settings.API_DOC_TITLE,
+      default_version=settings.API_DOC_DEFAULT_VER,
+      description=settings.API_DOC_DESCRIPTION,
+      terms_of_service="https://www.google.com/policies/terms/",
```
mrawls, line 30: Is this just a placeholder?

trail/upload/serializers.py
```diff
@@ -0,0 +1,31 @@
+from .models import Metadata
+from rest_framework import serializers
+
+
+class DynamicMetadataSerializer(serializers.ModelSerializer):
+    """
+    A ModelSerializer that takes an additional `fields` argument that
+    controls which fields should be displayed.
```
mrawls, line 8: Best docstring ever 💯 please make the rest like this, a nice clear and complete sentence! 😎

mrawls (commented):
Most of my comments are requesting clearer docstrings and more consistent docstring formatting. I suggest following this guide for function and class docstrings as much as feasible, both for content and format. It is based on numpydoc, and has easy-to-follow examples. https://developer.lsst.io/python/numpydoc.html

mrawls:
I defer to @DinoBektesevic for final merge approval on this PR

trail/trail/urls.py
didntpay: Can you let me know which part you are talking about?
For the settings.py I will bring it up in the meeting and see if there are better solutions

trail/trail/settings.py
didntpay: What would be a good solution for this, this is for drf_yasg indeed.

didntpay: I could move this part directly into the configs part for drf-yasg but that would require me to declare a new config.fromYaml there

trail/trail/urls.py
mrawls: The terms of service URL in particular

didntpay: I think I got it from Dino. @DinoBektesevic correct me if I am wrong

trail/config/site.yaml
DinoBektesevic: This would require me to set up admin emails and functionality so I think we can't get this done in this PR. Best if we leave empty or put one of our .uw.edu emails.

trail/query/tests.py
```diff
@@ -1,3 +1,92 @@
...
+
+    def testBasicWcsQuery(self):
+        """Tests getMeatadataInSpecifiedSky in metadataDao is functional"""
+        queryParam = {"raLow": 0, "raHigh": 1000, "decLow": 0, "decHigh": 1000}
+        response = self.metadataDao.queryBySpecifiedSky(queryParam)
+
+        self.assertEqual(len(response), 1)
+
+    def testCoordinateConversion(self):
+        ra = 200 * np.pi / 180
```
DinoBektesevic, line 81: Could live with a simple one-line comment. Could be its own test.

trail/query/views.py
```diff
@@ -61,3 +66,226 @@ def print_results(request):
...
+    DECHIGH = "decHigh"
+
+    def queryByParams(self, paramDict):
+
+        """
+        Returns the Metadata entries that meets the users input
+
+        Parameters
+        ----------
+        queryParam: dictionary
```
DinoBektesevic, line 89: use `` `dict` ``

trail/query/views.py
```diff
@@ -61,3 +66,226 @@ def print_results(request):
...
+            List of criterias for metadata
+
+        Returns
+        -------
+        results: list of metadata objects
```
DinoBektesevic, line 94: `` `list[Metadata]` ``

trail/query/views.py
```diff
@@ -61,3 +66,226 @@ def print_results(request):
...
+        """
+
+        querySet = Metadata.objects.all()
+        if paramDict is not None:
+            querySet = querySet.filter(**paramDict)
+
+        return list(querySet)
+
+    def queryBySpecifiedSky(self, paramDict):
+
```
DinoBektesevic, line 109: No new line

trail/query/views.py
DinoBektesevic: I agree the params should be more like

```
Parameters
----------------
paramDict: `dict`
    Dictionary of parameters. Must contain the following 
    keys: `raLow`, `raHigh`, `decLow`, `decHigh`
```

or document more details in the `Notes`

trail/query/views.py
```diff
@@ -61,3 +66,226 @@ def print_results(request):
...
+                "wcs_center_z__gte": lowerRight["z"]
+                }
+
+    def isWcsQueryParamMissing(self, queryParams):
+        """
+        Check if the sky boundary is correctly specified
+
+        Parameters
+        ----------
+        queryParams: Dictionary
```
DinoBektesevic, line 195: `` `dict` ``

trail/query/views.py
```diff
@@ -61,3 +66,226 @@ def print_results(request):
...
+    Support Query that returns metadata entry as a result through metadata info
+
+    Attributes
+    ----------
+    metadataDao: MetadataDao
+        Data accessing objects for metadata table
+
+    """
+
+    DATA_FIELDS = "fields"
```
DinoBektesevic, line 218: Inline a explanation of the class attribute like:

```
DATA_FIELDS = "fields"
"""Name of the key that designates returned Metadata fields."""

File: trail/trail/settings.py (hunk 9 above)
Comment (DinoBektesevic) [reply trail/trail/settings.py:53]: The thing that confused me is that I could not find any results that must include the expression `API_DOC_TITLE` when I search any combination of `drf-yasg django "API_DOC_TITLE"`

But what really happens here is that you declare this here so that in the `urls.py` you can call on these settings to set the schema_view attributes. Now, if we could have a `OPENAPI_SETTINGS` = config.toDict()` here and then in `urls.py` we can do `get_schema_view( **OPEN_API_SETTINGS)` that would be swell. Unfortunately half of this setup is being cast into objects like `openapi.Contact` or `openapi.License` making this impossible. 

So leave it here as is because the OpenAPI interface is really too convoluted to deal with in this PR.

File: trail/trail/urls.py (hunk 29 above)
Comment (DinoBektesevic) [reply trail/trail/urls.py:30]: I don't know why we would be linking to Google Account policies. If anything it should probably be: https://github.com/dirac-institute/trailblazer/blob/main/LICENSE

File: trail/trail/urls.py [hunk 37]
```diff
@@ -18,11 +18,31 @@
...
+      contact=openapi.Contact(email=settings.API_DOC_CONTACT_EMAIL),
+      license=openapi.License(
+          name=settings.API_DOC_LICENSE_NAME,
+          url=settings.API_DOC_LICENSE_URL
+        ),
+   ),
+   public=True,
+   permission_classes=[permissions.AllowAny],
```
DinoBektesevic, line 38: Should we make this `ReadOnly` for now?

trail/upload/models.py
```diff
@@ -205,7 +205,7 @@ class Wcs(models.Model, StandardizedKeysMixin):
     """Standardized keys that are tested only in approximate equality."""
 
     # same as above, cascading can orphan WCS entries
-    metadata = models.ForeignKey(Metadata, on_delete=models.PROTECT)
+    metadata = models.ForeignKey(Metadata, related_name='metadata', on_delete=models.PROTECT)
```
DinoBektesevic, line 208: Why this? Shouldn't it figure it out by itself?

trail/query/views.py
```diff
@@ -61,3 +66,226 @@ def print_results(request):
...
+        Return
+        ------
+        Return true if any of the param needed is not there
+        """
+        return not (self.RALOW in queryParams and self.RAHIGH in queryParams
+                    and self.DECHIGH in queryParams and self.DECLOW in queryParams)
+
+
+class MetadataQuery(APIView):
+
```
DinoBektesevic, line 207: Do we need the newline here?

trail/query/views.py
```diff
@@ -61,3 +66,226 @@ def print_results(request):
...
                   {"data": query_results, "wcsdata": wcs_list, "queryform": form, "render_table": True})
+
+
+class MetadataDAO(APIView):
+    """
+     Data Accessing objects for metadata, all query for metadatas are implemented here
+
+    """
+
+    RALOW = "raLow"
```
DinoBektesevic, line 77: Inline simple explanations of the class attributes. This also gets documented.

```
RALOW = "raLow"
"""Sets the name of the key in the request under which we expect to find the lower right ascension limit used in sky area queries."""
```

and same for others.

trail/query/views.py
```diff
@@ -61,3 +66,226 @@ def print_results(request):
...
+        Returns
+        -------
+        results: list of metadata objects
+            a list of metadata objects that matches the param specified by users.
+
+        Note
+        -----
+        If users do not specify a dict, the method will return all metadats
+        """
+
```
DinoBektesevic, line 101: Sometimes you leave a newline after docstrings and sometimes you do not. [PEP-257](https://peps.python.org/pep-0257/) recommends no newlines after.

trail/query/views.py
```diff
@@ -61,3 +66,226 @@ def print_results(request):
...
+        ----------
+        metadaIds: list
+            A list of metadata id that we need to fetch
+        """
+        result = []
+        for id in metadataIds:
+            result.append(Metadata.objects.get(id=id))
+        return result
+
+    def makeFilterDictForWcs(self, upperLeft, lowerRight):
```
DinoBektesevic, line 163: Perhaps something like `createWcsQueryDict` instead because `Filter` is overloaded term (django, metadata and then here too).

trail/query/views.py
```diff
@@ -61,3 +66,226 @@ def print_results(request):
...
+        """
+        Helper function for putting the sky boundary into a dictionary
+        that could be pass into querySet.filter()
+
+        Parameters
+        ----------
+        upperLeft: dictonary
```
DinoBektesevic, line 170: `` `dict` ``

trail/query/views.py
```diff
@@ -61,3 +66,226 @@ def print_results(request):
...
+            The upper left boundary for the sky specified
+        lowerRight: dictionary
```
DinoBektesevic, line 172: `` `dict` ``

trail/query/views.py
```diff
@@ -61,3 +66,226 @@ def print_results(request):
...
+
+    DATA_FIELDS = "fields"
+
+    def __init__(self):
+        self.metadataDao = MetadataDAO()
+
+    # gets all the metadata entry that matches the query parameters.
+    @swagger_auto_schema(responses={200: MetadataSerializer(many=True)})
+    def get(self, request):
+
```
DinoBektesevic, line 226: Remove newline

trail/query/views.py
```diff
@@ -61,3 +66,226 @@ def print_results(request):
...
+        """
+        Returns the Metadata entries that meets the users input in json
+
+        Parameters
+        ----------
+        fields: List of strings that is the key of wanted string
+            list of wanted field in the metadata returned
+        queryParam: Dictionary
+            list of criterias for metadatas specified by users
```
DinoBektesevic, line 235:
```
Parameters
---------------
request: `rest_framework.request.Request`
    Request containing the query values and return fields.  
```

If there are particular keys request has to hold and they don't fit in the Parameters use notes, otherwise I thought this is why we have yasg? In any case you can add:

```
Notes
--------
A request may contain the key `field` which is a list of `Metadata` field names 
that will be returned. See `Metadata` model for the list of allowed keys.
When not given, all `Metadata` fields will be returned.

A request may contain the key `queryParam` which is a dictionary of `Metadata`
field names. A subset of all `Metadata` entries that match the provided values, 
under the given keys, will be returned. If no `queryParams` are provided all 
`Metadata` entries are returned.
See `Metadata` model for a list of allowed keys.
```

Perhaps supplemented by an example section making it very obvious.

trail/query/views.py
```diff
@@ -61,3 +66,226 @@ def print_results(request):
...
+
+        Return
+        ------
+        A json response that contains the list of metadatas wanted
```
DinoBektesevic, line 239:
```
Returns
-----------
result: `rest_framework.response.Response`
    A JSON response containing matching `Metadata` entries.

File: trail/query/views.py [hunk 48]
```diff
@@ -61,3 +66,226 @@ def print_results(request):
...
+    """
+    Support Query that returns metadata entry that is in the part of sky specified by the users.
+
+    Attributes
+    ----------
+    metadataDao: MetadataDao
+        Data accessing objects for metadata table
+
+    """
+
```
DinoBektesevic, line 265: No new line

trail/query/views.py
```diff
@@ -61,3 +66,226 @@ def print_results(request):
...
+        """
+        Returns the Metadata associated with the part of sky the user specifies
+
+        Parameters
+        ----------------
+        raLow: the lower bound for ra
+        raHigh: the upper bound for ra
+        decLow: the lower bound for dec
+        decHigh: the upper bound for dec
```
DinoBektesevic, line 279: Parameters are things the function actually takes in. In this case:

````
Parameters
---------------
request: `rest_framework.request.Request`
    Request object containing the query parameters which must contain:
    `raLow`, `raHigh`, `decLow` and `decHigh` keys; the values of which 
    set the sky bounding box limits.
```

Also missing a `Returns` section

```
Returns
-----------
metadata : `rest_framework.response.Response`
    JSON response containing `Metadata` entries within the specified
   bounding box.
```

trail/upload/serializers.py
```diff
@@ -0,0 +1,34 @@
...
+    """
```
DinoBektesevic, line 9: This could use an Arguments section because otherwise nobody knows what key we need to specify the return fields under and the general signature of init means nothing will be documented.

```
Arguments
---------------
fields : `list[str]`
    List of keys that will be serialized into a response.

File: trail/upload/serializers.py [hunk 51]
```diff
@@ -0,0 +1,34 @@
...
+            existing = set(self.fields)
+            for field_name in existing - allowed:
+                self.fields.pop(field_name)
+
+
+class MetadataSerializer(DynamicMetadataSerializer):
+    """
+    A MetadataSerializer that takes an additional `fields` argument that
+    controls which fields should be displayed.
+    """
```
DinoBektesevic: Same as above. Remember, we inherit the init, except in this case we can also direct users to which key names are allowed as well.

```
Arguments
---------------
fields :  `list[str]`
    List of `Metadata` fields that will be serialized in the response.
```

trail/query/views.py
```diff
@@ -61,3 +66,226 @@ def print_results(request):
...
+    metadataDao: MetadataDao
+        Data accessing objects for metadata table
+
+    """
+
+    def __init__(self):
+        self.metadataDao = MetadataDAO()
+
+    def get(self, request):
+
```
DinoBektesevic, line 270: No new line.

trail/query/views.py
```diff
@@ -61,3 +66,226 @@ def print_results(request):
...
+            the lower bound for dec
+        decHigh: float
+            the upper bound for dec
+
+        Returns
+        -------
+        metadatas: list
+            list of metadatas that is in the part of sky specified by the users
+
+        Note
```
DinoBektesevic: `Notes`

trail/query/views.py
```diff
@@ -61,3 +66,226 @@ def print_results(request):
...
+        -------------------
+        if the parameters is not specified correctly, will return an empty result
```
DinoBektesevic: This is not true. You return an empty list.

DinoBektesevic (approved):
A lot of comments, nearly all related to the documentation. Only one or two are actual functionality questions. 

Well done, thank you for all the effort!

trail/query/views.py
didntpay: I think it is. Since we already have the objects which is what we need to return i just converted it into a list and return.

PR #47, opened 2022-04-24
GMOSN Standardizer

arihani2:
Implemented a standardized for images from the Gemini North Telescope. When passing images in, the astrometry.net key error is still thrown as expected, but images can be processed by astrometry.net.

### DIFF
--- a/trail/upload/process_uploads/standardizers/gmosn_standardizer.py
+++ b/trail/upload/process_uploads/standardizers/gmosn_standardizer.py
@@ -42,6 +42,7 @@ class GmosnStandardizer(HeaderStandardizer):
 
         # TODO: filter out what is the filter standardization here?
         # After uploading, no "astrometry.net key" runtime err?
+        # Do we need science program?
         meta = Metadata(
             obs_lon=self.header["GEOLON"],
             obs_lat=self.header["GEOLAT"],
@@ -49,11 +50,11 @@ class GmosnStandardizer(HeaderStandardizer):
             datetime_begin=begin,
             datetime_end= end,
             telescope="Gemini North",
-            instrument="GMOS", 
+            instrument="Hamamatsu CCD", #Instrument given from official
             exposure_duration=self.header["EXPTIME"],
             filter_name=self.header["FILTER"].strip()
         )
-
+        
         return meta
 
     def standardizeWcs(self, **kwargs):
--- a/trail/upload/process_uploads/standardizers/gmosn_standardizer.py
+++ b/trail/upload/process_uploads/standardizers/gmosn_standardizer.py
@@ -2,7 +2,7 @@
 Class that facilitates header metadata translation for Gemini North instrument
 """
 
-from datetime import datetime, timedelta
+from datetime import datetime, timedelta, timezone
 
 from astropy.time import Time
 
@@ -16,7 +16,7 @@ __all__ = ["GmosnStandardizer", ]
 class GmosnStandardizer(HeaderStandardizer):
 
     name = "gmosn_standardizer"
-    priority = 2
+    priority = 1
 
     def __init__(self, header, **kwargs):
         super().__init__(header, **kwargs)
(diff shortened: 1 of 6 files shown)


trail/upload/process_uploads/standardizers/gmosn_standardizer.py
```diff
@@ -0,0 +1,67 @@
...
+from upload.models import Metadata, Wcs
+
+
+__all__ = ["GmosnStandardizer", ]
+
+
+class GmosnStandardizer(HeaderStandardizer):
+
+    name = "gmosn_standardizer"
+    priority = 2
```
DinoBektesevic: Priority 1 please.

trail/upload/process_uploads/standardizers/gmosn_standardizer.py
```diff
@@ -0,0 +1,67 @@
...
+        # After uploading, no "astrometry.net key" runtime err?
+        # Do we need science program?
+        meta = Metadata(
+            obs_lon=self.header["GEOLON"],
+            obs_lat=self.header["GEOLAT"],
+            obs_height= 4213, #height in meters from official website
+            datetime_begin=begin,
+            datetime_end= end,
+            telescope="Gemini North",
+            instrument="Hamamatsu CCD", #Instrument given from official
```
DinoBektesevic: I don't think this is necessarily true across the board. We have the approximate dates where we can assign the original instruments to data so a couple of if-else questions to determine this would be better.

trail/upload/process_uploads/standardizers/gmosn_standardizer.py
```diff
@@ -0,0 +1,67 @@
...
+            exposure_duration=self.header["EXPTIME"],
+            filter_name=self.header["FILTER"].strip()
+        )
+        
+        return meta
+
+    def standardizeWcs(self, **kwargs):
+        # ignores any WCS information in the header and instead just sends it to astrometry.net to solve.
```
DinoBektesevic: Hmm. Can you elaborate on why do we need to force this to go to Astrometry.NET? 

I wanted the default logic flow to be 1) try to read the header for WCS --> if it fails: send to astrometry.net; else convert. This is how it's currently written [here](https://github.com/dirac-institute/trailblazer/blob/main/trail/upload/process_uploads/header_standardizer.py#L314). I see MOA has the same standardizer method, but are these really necessary or did you just cargo-cult it over?

If it's necessary then we should fix the code, but if it was cargo-culted then we can delete these methods because that's what's going to happen anyhow.

DinoBektesevic (approved):
If you fix flake8 errors and the tests pass we can merge this PR.

trail/upload/process_uploads/standardizers/gmosn_standardizer.py
```diff
@@ -0,0 +1,65 @@
...
+        return False
+
+    def standardizeMetadata(self):
+        DATEOBS = self.header["DATE-OBS"]
+        EXP = self.header["EXPTIME"]
+        begin = datetime.strptime(DATEOBS, "%Y-%m-%dT%H:%M:%S.%f")
+        begin = begin.replace(tzinfo=timezone.utc)
+        end = begin + timedelta(seconds=EXP)
+
+        # Need confirmation on what october date. No clear day,
```
mrawls: Can you please explain what's happening here? If observations are older than Oct 2011, I get you're doing one thing, and if they're older than June 2014, you're doing another thing. What's the motivation for this? A one-liner comment explaining would go a long way.

trail/upload/process_uploads/standardizers/gmosn_standardizer.py
```diff
@@ -0,0 +1,65 @@
...
+        # but says mid october. Also not sure what day in June,
+        # assumed first day for now
+        oct_string = "2011-10-01T00:00:00.0"
+        october11 = datetime.strptime(oct_string, "%Y-%m-%dT%H:%M:%S.%f")
+        october11 = october11.replace(tzinfo=timezone.utc)
+        june_string = "2014-06-01T00:00:00.0"
+        june14 = datetime.strptime(june_string, "%Y-%m-%dT%H:%M:%S.%f")
+        june14 = june14.replace(tzinfo=timezone.utc)
+
+        if begin < october11:
```
mrawls: If an observation falls between the two dates, your if/else doesn't seem to handle this properly.

trail/upload/process_uploads/standardizers/gmosn_standardizer.py
```diff
@@ -0,0 +1,65 @@
...
+Class that facilitates header metadata translation for Gemini North instrument
+"""
+
+from datetime import datetime, timedelta, timezone
+
+from upload.process_uploads.header_standardizer import HeaderStandardizer
+from upload.models import Metadata
+
+
+__all__ = ["GmosnStandardizer", ]
```
mrawls: I suggest a clearer name here, such as "GeminiNorthStandardizer" (and, correspondingly, `.geminiNorth_standardizer` etc.) This is just for Gemini North and not South, correct? Long variable names are fine if they add clarity.

mrawls (commented):
If you can address my comments and get rid of the flake8 errors, this looks great to me.

If you're using VSCode, you can pretty easily use flake8 by following these directions https://code.visualstudio.com/docs/python/linting. It's helpful to run it yourself before pushing to GitHub, so the GitHub check is verifying the syntax is good vs catching all the initial syntax errors (and yes, everybody has loads of them on a first pass, it's not just you 😄 )

trail/upload/process_uploads/standardizers/gmosn_standardizer.py
DinoBektesevic: GMOSN instruments changed throughout the years and were replaced with newer CCDs. Turns out their website has  this data written down to a month-like precision and while I suspected we would not be getting many (any) old data as that we might as well put that logic in. So what this is doing here is checking the image date against the dates on which they reported instrument changes, but given changing out the instrument is not exactly a 5 minute deal, he made some educated guesses.

It's not the best, but I told him it'll be good enough given the circumstances. A comment would be good.

mrawls: OK! I'm fine with the educated guesses, just wanted a comment better explaining the situation.

arihani2:
[Add functionality for Gemini North Telescope](https://trello.com/c/BM0b0zBD/48-add-functionality-for-gemini-north-telescope)

trail/upload/process_uploads/standardizers/gmosn_standardizer.py
mrawls: Nevermind, this uses `elif` and not `if`, and the logic is fine, I just misread it!

PR #50, opened 2022-06-01
Lowell Standardizer

arihani2:


arihani2:
[Lowell Discovery Telescope Standardizer](https://trello.com/c/r9P4LviY/52-lowell-discovery-telescope-standardizer)

mrawls (approved):
This looks excellent! Great job on the docstrings in particular. One question, I see you added a test image and some standardized values to the tests directory, but I don't think I see anywhere that a test is actually run. Is that happening elsewhere/later?

arihani2:
> I don't think I see anywhere that a test is actually run. Is that happening elsewhere/later?

The tests run automatically when you add the cutout file to the upload/data/tests directory and when you add the expected values to the yaml file. The tests go through all the files in the directory so no new test needs to be written.

trail/upload/process_uploads/standardizers/lowell_standardizer.py
```diff
@@ -0,0 +1,72 @@
+
```
DinoBektesevic: No need for this newline

DinoBektesevic (approved):
Looks good!

PR #51, opened 2022-06-09
Long overdue overhaul of website text

mrawls:
Finally added content to the GitHub README and the website About page!

I tweaked some CSS formatting defaults and text on the other landing pages, too.

Dino taught me that when editing `main.css`, one must kill the server, do `python manage.py collectstatic`, and then once again do `python manage.py runserver` to see the changes. 🎉


trail/gallery/templates/gallery_table.html
```diff
@@ -18,5 +18,4 @@
 <!-- Footer -->
 <footer>
     {% include "gallery_pagination.html" %}
-</div>
```
DinoBektesevic, line 21: This random floating div closed the `wrapper` element and prevented the background to extend all the way to the end of page.

trail/trail/templates/about.html
```diff
@@ -445,8 +525,9 @@ <h3>Social</h3>
 				</section>
 		</footer>
 
-		<!-- Copyright -->
+
 		<div id="copyright">
 				<ul><li>&copy; Untitled</li><li>Design: <a href="https://html5up.net">HTML5 UP</a></li></ul>
 		</div>
+-->
```
DinoBektesevic: This comment contains nearly 18 000 characters, which is nearly as big as the plot on this page (~3 times the text). I'm not in favor of keeping it in this file for many reasons, download size and SEO being up there.

If you want to keep it as a reference I would separate it from the actual HTML that is parsed and downloaded on every page visit and put it in a different file. I'm ok if that file lives in static directory.

trail/trail/templates/base.html
```diff
@@ -59,15 +57,16 @@
...
+						    <!-- <section>
+								<h3>Social</h3>
 								    <ul class="icons alt">
 									      <li><a href="#" class="icon brands alt fa-twitter"><span class="label">Twitter</span></a></li>
 									      <li><a href="#" class="icon brands alt fa-facebook-f"><span class="label">Facebook</span></a></li>
 									      <li><a href="#" class="icon brands alt fa-instagram"><span class="label">Instagram</span></a></li>
 									      <li><a href="https://github.com/dirac-institute/trailblazer" class="icon brands alt fa-github"><span class="label">GitHub</span></a></li>
 								    </ul>
 						    </section>
+                            -->
```
DinoBektesevic: If you like an empty `<section>` because it gives a gray bck it should be safe to return it now (the background fix was an unpair closing div element, see above), but I would delete the content instead of commenting it out.

trail/upload/templates/upload.html
```diff
@@ -3,10 +3,24 @@
...
+
 {% block content %}
+<!-- Post -->
+<section class="post">
+    <header class="major">
+            <h1>Upload</h1>
+            <p>We welcome contributions of satellite-streaked FITS files to Trailblazer.<br />
+               Reduced and photometrically calibrated images are preferred, but not required.<br />
+               By uploading, you agree the files will be used in accordance with a
+               <a href="https://creativecommons.org/licenses/by/4.0/">CC-BY</a> license.
```
DinoBektesevic, line 19: I'm fine with the comment here but a heads up, just in case, that they will have to click on a checkbox in the upload form also consenting to this so the notification is duplicated.

trail/trail/templates/about.html
```diff
@@ -3,18 +3,103 @@
...
+		<p>A rapidly increasing population of commercial low-Earth orbit (LEO) satellites is beginning to impact ground-based astronomy.
+			LEO satellites reflect sunlight and cause bright streaks in telescope images.
+			For an overview of the situation and its impacts to optical astronomy and beyond, see <a href=https://ui.adsabs.harvard.edu/abs/2022NatAs...6..428L/abstract>Lawrence et al. 2022</a>.<br /><br />
+		<figure>
+			<span class="image left">
+				<img src="../static/images/lawrence2022_fig1.png" alt="Plot showing increasing numbers of tracked objects in orbit from 1957 to 2021." />
+				<figcaption><br />Growth of all tracked objects in orbital space over time.
+					“Dead” is derelict satellites, “Leftovers” is parts such as rocket stages, and “Debris” is smaller tracked material.
+					Commercial satellite constellations appear in the rapid “Active” line increase, and are much brighter than typical leftovers or debris.
+					Figure 1 from Lawrence et al. 2022.</figcaption>
```
DinoBektesevic: I feel like we can make this element `<figcaption style="font-size: smaller;">` which will set the font responsively to a smaller than the main text.

trail/trail/templates/about.html
```diff
@@ -3,18 +3,103 @@
...
 <section class="post">
+		<!--
 		<header class="major">
-				<h1>This is<br />
-						all about me</h1>
+				<h1>About Trailblazer</h1>
 		</header>
+		-->
+
+		<h2>The problem &#128752;</h2>
```
DinoBektesevic: I think we should make all our titles anchor points too. Something like:
```
<h2>
  <a name="our-approach"></a>
  <a class="headerlink" href="#our-approach" title="Permalink to this headline">Our approach &#129321;</a>
</h2>
```

Then people could link directly to a section like `/about/#our-approach` which I just found always very practical.

DinoBektesevic (approved):
Looks good, finally some actual content for the frontend.

trail/upload/templates/upload.html
mrawls: I'd forgotten about that! I feel like duplication is OK here though 👍

trail/trail/templates/about.html
mrawls: I knew that was a thing, but wasn't sure how - thanks

trail/gallery/templates/gallery_table.html
mrawls: Thanks for fixing it!

PR #52, opened 2022-06-09
LBT Processor

arihani2:


arihani2:
[Create Processor for LBT](https://trello.com/c/r2tlIcpX/51-create-processor-for-lbt)

trail/upload/process_uploads/fits_processor.py
```diff
@@ -331,8 +331,8 @@ def process(self):
         else:
             standardizedResult.extendThumbnails(tmpResult)
 
-        if len(standardizedResult.thumbnails) != len(standardizedResult.wcs):
-            raise RuntimeError("Can not unambiguously assign thumbnails to WCSs!")
+        # if len(standardizedResult.thumbnails) != len(standardizedResult.wcs):
```
DinoBektesevic: Delete these lines. The number of thumbnails does not need to match the number of WCS results in general. I would open a ticket (GH issue or trello) mentioning we don't test with DECAM Community pipeline products and that's why we missed this and that it might be good if we did that.

trail/upload/process_uploads/fits_processor.py
```diff
@@ -331,8 +331,8 @@ def process(self):
...
+        #     raise RuntimeError("Can not unambiguously assign thumbnails to WCSs!")
 
         for thumb, wcs in zip(standardizedResult.thumbnails, standardizedResult.wcs):
```
DinoBektesevic: This is then also not correct. We would have to check if `thumbnails` is a list or not and if it is then it must be the same length, otherwise we need to go and assign each wcs the same `thumbnails` because they are all pasted in a single image.

trail/upload/process_uploads/processors/lbt_processor.py
```diff
@@ -0,0 +1,102 @@
...
+
+    def createThumbnails(self, scaling=(4, 10)):
+
+        # NAXIS1  =                 2304 / Image x size
+        # NAXIS2  =                 4608 / Image y size
+        # Center of image: 3,466
+
+        padding = 10
+        xdim = 2304*3 + 2*padding
+        ydim = 4608 + padding + 2304
```
DinoBektesevic: It would be nice, for readability sake that variables such as these become either a) global variables in this module or b) put into a class called detector, or c) some other option down these lines. For example, if we had:

```
class Instrument:
    img_dimx = 2304
    img_dimy = 4608
    focal_plane_dimx= 2304*3 + 2*padding
    focal_plane_dimy= 4608 + padding + 2304
    etc...
```
then in later code these don't look like magic numbers pulled out of nowhere but 
```
img[- Instrument.img_dimx:,  -Instrument.img_dimy:] = ccddata
```
See for example `decam_processor`. This would drastically improve code readability.

trail/upload/process_uploads/processors/lbt_processor.py
```diff
@@ -0,0 +1,102 @@
...
+            image = self.normalizeImage(image)
+
+            # TODO: test here if the step-vise resizing is faster...
+            image = image.resize((focalPlane.scaledY, focalPlane.scaledX),
+                                 Image.ANTIALIAS)
+
+            focalPlane.add_image(image, ext.header["DETPOS"])
+
+        return focalPlane
+
```
DinoBektesevic: This is the DECam processor method we copied here for reference. It isn't used anywhere in the code so remove it.

trail/upload/process_uploads/processors/lbt_processor.py
```diff
@@ -0,0 +1,102 @@
...
+    def createThumbnails(self, scaling=(4, 10)):
+
+        # NAXIS1  =                 2304 / Image x size
+        # NAXIS2  =                 4608 / Image y size
+        # Center of image: 3,466
+
+        padding = 10
```
DinoBektesevic: It would be nice if this were a adjustable class or, probably better, instance attribute.

trail/upload/process_uploads/processors/lbt_processor.py
```diff
@@ -0,0 +1,102 @@
...
+
+            focalPlane.add_image(image, ext.header["DETPOS"])
+
+        return focalPlane
+
+    def createThumbnails(self, scaling=(4, 10)):
+
+        # NAXIS1  =                 2304 / Image x size
+        # NAXIS2  =                 4608 / Image y size
+        # Center of image: 3,466
```
DinoBektesevic: These constants that refer to the detector should probably not be hardcoded into this method but extracted into either module variables or into a class where they can be documented properly, in which case this comment won't probably be necessary anymore.

trail/upload/process_uploads/processors/lbt_processor.py
```diff
@@ -0,0 +1,102 @@
...
+
+            # TODO: test here if the step-vise resizing is faster...
+            image = image.resize((focalPlane.scaledY, focalPlane.scaledX),
+                                 Image.ANTIALIAS)
+
+            focalPlane.add_image(image, ext.header["DETPOS"])
+
+        return focalPlane
+
+    def createThumbnails(self, scaling=(4, 10)):
```
DinoBektesevic: You declare scaling as function parameter but don't use it anywhere? Also are these appropriate scaling values? I would check that by looking at what the produced image looks like in the gallery and balancing the quality with the size. Note that in the gallery the more important bit is that the images are as small as possible, while the large image can be substantially bigger, but probably shouldn't really break 5MB in size tops (I would expect ~2MB off the top of my head).

trail/upload/process_uploads/processors/lbt_processor.py
```diff
@@ -0,0 +1,102 @@
...
+        for ccd in self.exts:
+            ccddata = self.normalizeImage(ccd.data)
+            if '1' in ccd.header['EXTNAME']:
+                img[-2304:, -4608:] = np.fliplr(ccddata.T)
+            elif '2' in ccd.header['EXTNAME']:
+                img[2304+padding:2*2304+padding, -4608:] = np.fliplr(ccddata.T)
+            elif '3' in ccd.header['EXTNAME']:
+                img[:2304, -4608:] = np.fliplr(ccddata.T)
+            else:
+                img[1162:1162+4608, :2304] = np.flip(ccddata)
```
DinoBektesevic: A comment on how we landed on these numbers might be helpful. Even better would be if you calculated this as a module global var, or instance variable of a detector or something.

trail/upload/process_uploads/processors/lbt_processor.py
```diff
@@ -0,0 +1,102 @@
...
+
+        img = (img - img.min()) * 255/(img.max() - img.min())
+        img = Image.fromarray(img.astype(np.uint8), 'L')
+
+        newY, newX = int(ydim/2), int(xdim/2)
```
DinoBektesevic: Scaling factors are hardcoded even though they exist as parameters to the function

trail/upload/process_uploads/processors/lbt_processor.py
```diff
@@ -0,0 +1,102 @@
...
+        img = img.resize((newY, newX), Image.ANTIALIAS)
+        newY, newX = int(ydim/4), int(xdim/4)
```
DinoBektesevic: Scaling factors again.

DinoBektesevic (changes requested):

DinoBektesevic (approved):

PR #53, opened 2022-06-10
VATT Standardizer

arihani2:


arihani2:
[Add support for Vatican Observatory VATT (create new standardizer)](https://trello.com/c/gf3ikOb9/57-add-support-for-vatican-observatory-vatt-create-new-standardizer)

DinoBektesevic (approved):
I think this looks good but the size of the cutout files for the tests is too big. If you replace that with a smaller one (see for example the sizes of the other cutouts in the `upload/test/data` directory, all are smaller than 1MB) you can go ahead and merge.

PR #56, opened 2022-10-21
Add Whipple Observatory header standardizer.

DinoBektesevic:
The main problem with Whipple observatory files is that they are smaller than 1MB, which made Django never write them to temporary file, which made AstroPy fail to open them. Second issue was that I didn't account for lower/uppercase file extensions.

I went ahead and added a simple header standardizer for Whipple. It does not cover all of the telescopes at that site. This should be fixed/expanded on in the future. 

An erroneous commit https://github.com/dirac-institute/trailblazer/commit/e960757c7f2c61f97924c24c911c16c4bf93392a commented out the default ASTROMETRY.NET setting. This wasn't caught because the deployed website settings file diverged and didn't have this issue and nobody manually tested if upload to astrometry.net worked.

### DIFF
--- a/trail/upload/process_uploads/standardizers/whipple_standardizer.py
+++ b/trail/upload/process_uploads/standardizers/whipple_standardizer.py
@@ -37,7 +37,15 @@ class WhippleStandardizer(HeaderStandardizer):
         # (height specifically) and implement conditions to handle them
         # http://www.sao.arizona.edu/FLWO/whipple.html
         # This is the Ridge location height
-        height = 2340
+        height = self.header.get("HEIGHT", None)
+        if height is None:
+            if "cecilia" in self.header["TELESCOP"].lower() or "ben" in self.header["TELESCOP"]:
+                height = 1268.00
+            else:
+                raise RuntimeError(
+                    "Unable to parse height of the instrument. Header does "
+                    "not contain key HEIGHT or is not one of the supported "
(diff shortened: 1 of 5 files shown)


trail/upload/process_uploads/standardizers/whipple_standardizer.py
```diff
@@ -0,0 +1,55 @@
...
+
+        # TODO: there are multiple different telescopes
+        # at the Whipple Obs site, check if there are any particularities
+        # (height specifically) and implement conditions to handle them
+        # http://www.sao.arizona.edu/FLWO/whipple.html
+        # This is the Ridge location height
+        height = 2340
+
+        meta = Metadata(
+            obs_lon=self.header["LONGITUD"],
```
mrawls, line 51: Field is definitely truncated `LONGITUD` not `LONGITUDE`? Just checking

mrawls (approved):
Let's get this merged - I see your note about different Whipple elevations, but it looks like this would only matter for MMT, and I'm not immediately sure how to check if that is the telescope being used without a sample MMT FITS header.

trail/upload/process_uploads/standardizers/whipple_standardizer.py
DinoBektesevic: Yep, 8 char limit for fits header keys in general, I think, but def `LONGITUD` in this one.

